PR #1, opened 2016-08-12
KaitaiStruct: add a class method from_file()

antonv6:
Classes that inherit KaitaiStruct will then have this method create an instance
of the appropriate class. This can basically make the generated code smaller.


GreyCat:
Thanks!

PR #3, opened 2016-10-10
Use single python module instead of package

dgladkov:
Right now `setup.py install` does not work because current project layout is not actually a package as Python packages require `__init__.py` file inside project directory. Creating a package for one Python module and adding additional imports seems like an overkill, so I converted project to a simple module. I also cleaned up setup.py file because all properties besides classifiers are pretty much self explanatory. Setup works correctly now and everything is ready for publishing on PyPI.

If you need help publishing a package, feel free to send me a message (I speak Russian).


GreyCat:
I've put up some separate tests for Python 2 and Python 3 at our CI: http://kaitai.io/ci/
It looks like Python 3 support is pretty bad for now. Probably it's a good idea to fix it to acceptable level before publishing the package?

PR #9, opened 2016-12-03
Fix exception text and Python 2 compatibility improvements

athre0z:
The behavior of `class A` (without any explicit inheritance) changed between Python 2 and 3 from creating an old-style class to creating a new-style one. The first commit addresses this issue by explicitly inheriting `object`, always creating a new-style class. While the old behavior works, it makes it hard to write code that uses the classes (e.g. inherits them) that is compatible to Py 2 and 3 as well.

The second commit fixes the exception text formatting and changes the redundant parentheses into tuples (I suppose you probably wanted tuples and forgot the `,`, resulting in the parentheses having no effect).

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -189,7 +189,7 @@ class KaitaiStream(object):
         r = self._io.read(n)
         if len(r) < n:
             raise EOFError(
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks!

PR #10, opened 2017-03-15
Added __repr__ and a helper function

KOLANICH:


GreyCat:
Thanks for the contribution!

Unfortunately, this code breaks many things:

* First and foremost, it fails with an error:
```
     33 
     34     def __repr__(self):
---> 35         return "".join((self.__class__.__name__, "(", ", ".join(self._reprGeneratorForAllProps(self)), ")"))
     36 
     37     def __exit__(self, *args, **kwargs):

TypeError: _reprGeneratorForAllProps() takes exactly 1 argument (2 given)
```
* If it was working, this code dumps only `seq` items, but it will ignore instances, as they are stored in `_m_foo` like members, starting with `_`, which you skip
* Adding support for instances might be tricky, as you'll need to avoid problems with infinite recursion to be encountered in cases like these:
```yaml
meta:
  id: rec
seq:
  - id: some
    type: u1
instances:
  self_ref:
    value: _root
```
* Naming:
  * function names [must use lower underscore case](https://www.python.org/dev/peps/pep-0008/#function-names)
* Complexity:
  * cramming 4-5-6 statements into a single line is not a good idea
  * Iadding extra function in KaitaiStruct class that will be inherited everywhere is probably not a good idea as well
* Formatting:
  * PEP8 requires [usage of whitespace](https://www.python.org/dev/peps/pep-0008/#other-recommendations) around some binary operators and recommends to use it around the others
  * line length [must be up 79](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) — you have 109 here
  * docstring [must end with a period](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)
  * Codacy already reported trailing whitespace

kaitaistruct.py
```diff
@@ -27,6 +27,13 @@ def __init__(self, stream):
     def __enter__(self):
         return self
 
+    def _reprGeneratorForAllProps(self):
```
funkyfuture: this name doesn't comply with pep8.

arekbulski:
Forgive me for being outright dismissive but... is `__repr__` really that much needed? @GreyCat I noticed that you and I run our forums completely differently. In Construct issues, I would have closed this PR after no more than a day of deliberation and half the arguments. You seem to keep issues open forever for everyone to see and participate. I have to admit, and very ahem deliberately, that I appreciate the way you run things. But lets close this one and move on? :thinking:

KOLANICH:
__repr__ is definitely needed. It is a pain to deal (especially do debug in ipython shell) with anything without __repr__.

arekbulski:
I just fixed the style/pep8, didnt change how it works. Better?

KOLANICH:
I have added a guard against recursive objects.

KOLANICH:
Broadened scope to instances. @arekbulski @GreyCat could you test and review it, it seems it's ready.

kaitaistruct.py
```diff
@@ -15,6 +15,18 @@
 #
 __version__ = '0.8'
 
+from RecursionSafe import RecursionSafe
+
+recSafe=RecursionSafe()
```
funkyfuture: does flake8 not complain about that symbol name?

kaitaistruct.py
```diff
@@ -15,6 +15,18 @@
...
+
+@recSafe.wrap
+def repr_generator_for_all_props(self):
+  """Generator to use in own __repr__ functions."""
+  return (
+    "".join(( str(k), "=", repr(getattr(self, k)) ))
+    for k in dir(self)
```
funkyfuture: i don't know if it's usable in this context, but `vars` is often inferior over `dir` and might make test in the following line unnecessary.

kaitaistruct.py
```diff
@@ -15,6 +15,18 @@
...
+    if k[0] != "_" and not hasattr(KaitaiStruct, k) and not isinstance(getattr(self, k), type)
```
funkyfuture: `k.startswith('_')` would be more expressive.

kaitaistruct.py
```diff
@@ -26,6 +38,16 @@ def __enter__(self):
     def __exit__(self, *args, **kwargs):
         self.close()
 
+    def __repr__(self):
+        return "".join(
```
funkyfuture: these are too much line-breaks for my gusto.

setup.py
```diff
@@ -44,4 +44,6 @@
     ],
     keywords='kaitai,struct,construct,ksy,declarative,data structure,data format,file format,packet format,binary,parser,parsing,unpack,development',
     py_modules=["kaitaistruct"],
+    requires=["RecursionSafe"],
+    dependency_links=["git+https://github.com/KOLANICH/RecursionSafe.py.git"],
```
funkyfuture: i wouldn't trust a python library with camelcase naming and `.py` suffix. ;-)

GreyCat: Also:

> Requirements
> Python 3.

Which is, obviously, a deal-breaker for inclusion here anyway :(

> Python 2 is dead, stop raping its corpse.

Unfortunately, I see the exact opposite :(

kaitaistruct.py
KOLANICH: recSafe will be fixed to `rec_safe`. I don't use flake.

kaitaistruct.py
KOLANICH: `vars` doesn't return properties.

kaitaistruct.py
KOLANICH: It's longer and I'm not sure what is more performant.

arekbulski:
I dont even understand how this code works, nor why its so lengthy. Also its not providing different str and repr versions. I would suggest considering using Construct as reference:
- recursion lock
only 16 lines of code
https://github.com/construct/construct/blob/492b5f93679b6fb21f01558e20420efabc6dac01/construct/lib/containers.py#L29-L44
- both STR and REPR
REPR produces a full one-liner that entirely reproduces the dict.
STR produces newline-separated shortened items (truncated to 32 bytes) for viewing.
https://github.com/construct/construct/blob/492b5f93679b6fb21f01558e20420efabc6dac01/construct/lib/containers.py#L207-L250

arekbulski:
I appreciate the initiative, but honestly I dont like the implementation. If you would be willing to wait, then I will offer alternative implementation based on Construct.

KOLANICH:
There were some errors, I have fixed them.

KOLANICH:
reprlib is missing on python 2, but I guess may be backported and put into pip, the name is empty https://pypi.python.org/pypi/reprlib

arekbulski:
I have issues with this implementation:
- reprlib is Python 3 only but the runtime is supposed to be Python 2/3
- your code uses different indentation than rest of the script
- I dont particularly like the usage of dir, we could ask @GreyCat to change the compiler so that generated structs have something similar to `__slots__`, so you would just iterate that.
- it evalutates unevaluated (lazy) instances? this could lead to failure.

KOLANICH:
>reprlib is Python 3 only but the runtime is supposed to be Python 2/3

We can try to backport it.

>your code uses different indentation than rest of the script

fixed

>I dont particularly like the usage of dir, we could ask @GreyCat to change the compiler so that generated structs have something similar to __slots__, so you would just iterate that.

```__slots__``` cannot be properties

>it evalutates unevaluated (lazy) instances? this could lead to failure.

It does, it could. But reprlib provides recursion lock and truncation, so I guess there will be no infinite recursions and rendering too much values of sequences. About reading large files into memory which can cause consuming all the memory in machine - I guess we should map them and use references to them instead of copies (I wonder if it is possible in python), this would lead into consuming only a single page.

**Comment by arekbulski** [comment]
> ```__slots__``` cannot be properties

But thats not what I said or meant. We could have the compiler generate KaitaiStruct-s that have something similar to slots, a list of names, to iterate that instead of filtering dir:
```py
class Enum0(KaitaiStruct):
    __slotsalike__ = ["pet_1","pet_2"]
    def _read(self):
        self.pet_1 = self._root.Animal(self._io.read_u4le())
        self.pet_2 = self._root.Animal(self._io.read_u4le())
```
then
```py
-        for k in dir(self)
-        if k[0] != "_" and not hasattr(KaitaiStruct, k) and not isinstance(getattr(self, k), type)
+    for k in __slotsalike__: 
+    yield getattr(self, k)
```
This would not only have better performance, if you care about that in this case that is, but it might also be useful to end users as they might have a use for a list of field names.

arekbulski:
> But reprlib provides recursion lock and truncation,

That does not address the issue I meant. You could have a pointer instance that refers to outside of stream size, which will cause failure if evaluated. Its not about the size of the output.

KOLANICH:
>You could have a pointer instance that refers to outside of stream size, which will cause failure if evaluated. Its not about the size of the output.

Out of scope - its responsibility of a ksy developer to have all the needed `if`s.

>But thats not what I said or meant. We could have the compiler generate KaitaiStruct-s that have something similar to slots, a list of names, to iterate that instead of filtering dir:

I guess we need non-volatile fields in slots and properties in a separate variable. And tuples should be used, not lists.

GreyCat:
Compiler actually already does something similar when doing `--debug`, but not for Python. I have no idea what you guys mean by "non-volatile fields in `_fields`, properties in `_props` and everything possible in `__slots__`", i.e. what is "non-volatile fields", "properties", "everything possible".

We generate an array called something like "seq_fields", i.e.:

* https://github.com/kaitai-io/ci_targets/blob/master/compiled/ruby/debug_0.rb#L11
* https://github.com/kaitai-io/ci_targets/blob/master/compiled/java/src/io/kaitai/struct/testformats/Debug0.java#L22

Could you explain, what's wrong with current solution - i.e. is it's reflection slow, unreliable, or something else is the matter? As far as I have guessed, `__slots__` is some special array in Python?

KOLANICH:
>I have no idea what you guys mean by "non-volatile fields in _fields, properties in _props and everything possible in __slots__", i.e. what is "non-volatile fields", "properties", "everything possible".

Non-volatile fields are the fields (members in `seq`) without `volatile` hint. `_props` is a tuple of names of everything implemented as `@property`es: instances, volatile fields, etc. `__slots__` is a special collection listing the names of `dumb` (non-`@property` members), if it is present, only the names listed in it are allowed (unless you have some magic method overloading it) and it is claimed that memory footprint is reduced.

arekbulski:
[SEQ_FIELDS  ](https://github.com/kaitai-io/ci_targets/blob/master/compiled/ruby/debug_0.rb#L11)is exactly what I am talking about. The only issue with it is that it should be included even without debug mode and available in Python.

On Kolanich behalf: non-volatile is just a synonym for cached (like instances are cached).

z3ntu:
Hi, what's the state of this PR? I'm interest in this as well as the default string representation is not really useful at the moment `<MyBinaryType object at 0xb18605b0>` (and I wanted to spare myself writing it myself for a quick prototype :wink:)

KOLANICH:
The status of this PR is that I personally use it, but in a form of a separate branch on top of `master`. On large files it may cause problems, since text is generated first, then is output, we don't evaluate size of the text that will be generated before generating it.

PR #11, opened 2017-03-23
Note explicit support for Python 3.6

felixonmars:


GreyCat:
Thanks! Shall I release an update for 0.7 runtime here (i.e. because it fixes some dependency issues for Python 3.6 people), or it's purely cosmetical and can wait for 0.8?

PR #13, opened 2017-05-30
Refactored functions creation

KOLANICH:
performance issues from #12 should be solved now (use ```dis.dis(kaitaistruct.KaitaiStream.read_s4be)``` to understand why).

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -3,6 +3,14 @@ import sys
 from struct import unpack
 
 try:
+    from bind import bind
+except:
+    def bind(**kwargs):
+        def identity(f):
+            return f
+        return identity
+
+try:
     import cStringIO
     BytesIO = cStringIO.StringIO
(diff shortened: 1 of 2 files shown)


GreyCat:
This is unlikely to be accepted either:

* This still replaces simple and straightforward code with complex and harder-to-understand one. Basically any code metric (i.e. cyclomatic complexity, maintainability score, etc) indicates that it's more complex that it was.
* This brings extra dependency.
* You still haven't provided by benchmarks that support performance claims.
* Meta-programming machinery still potentially generates problems with using this code with Cython, Python-into-something transpilers, etc, etc.

KOLANICH:
I have done some benchmarking, it is really slower. Strangely, after using ```bind``` it became even slower (for now I don't know why, on synthetic tests bind gives speed-up, and the byte-code of optimized functions is identical to the original ones).

PR #17, opened 2018-01-18
cStringIO is ancient, modern python hides it

arekbulski:
cStringIO should not be imported, it was deprecated a long time ago. Python (since some modern version) hides that module and just exposes the BytesIO class. Its same speed on modern Python.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -1,12 +1,7 @@
 import itertools
 import sys
 from struct import unpack
-
-try:
-    import cStringIO
-    BytesIO = cStringIO.StringIO
-except ImportError:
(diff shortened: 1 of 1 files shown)


GreyCat:
We're actually in process of building some infrastructure that will run the code in multiple environments, you can see [work in progress](https://travis-ci.org/kaitai-io/ci_targets). Currently, it tries to run Python 2.7 and Python 3.6. If that's good enough for baseline support (i.e. you won't recommend supporting earlier Pythons), then it should be ok to switch, right?

arekbulski:
Construct supports 2.7+ and 3.3+ and it uses this code.

Issue #8, opened 2016-10-13
XOR benchmarks


dgladkov:
Followup of discussion in @04289cc13d6501153cab61e875d9c4e2ccdaa5d8

Python 2 bytearray mutations
```
== run_benchmark_process_xor.py
max = 2147483543
34.8218421936
0.414654970169
```

Python 3 bytearray mutations
```
== run_benchmark_process_xor.py
max = 2147483543
33.529802142000335
0.39478560099996685
```

Python 3 iterating bytes
```
== run_benchmark_process_xor.py
max = 2147483543
32.42236607199993
0.39769275300022855
```

Python 2 numpy bitwise_xor
```
== run_benchmark_process_xor.py
max = 2147483543
27.3349659443
0.424569129944
```

Python 3 numpy bitwise_xor
```
== run_benchmark_process_xor.py
max = 2147483543
24.56664780200026
0.4790448710000419
```

Conclusions:

* On real world example difference between bytearray mutations and map over bytes for Python 3 is not worth maintaining the separate code path.

* numpy's bitwise_xor gives significant performance increase, however it requires manual key wrapping as bitwise_xor expects arrays of the same size

As numpy contains C extensions and may require compilation, I suggest adding numpy as an optional dependency and falling back to native Python bytearray mutations + xor if numpy is not installed.

GreyCat:
Great comparison, thanks!

I have yet another humble suggestion - is it possible to submit something like `bitwise_xor_cyclic` to numpy, and then add here? That would be probably make it work even faster?

There's also this [large battle for XOR](http://stackoverflow.com/questions/2119761/simple-python-challenge-fastest-bitwise-xor-on-data-buffers) at SO. I can't estimate how hard is it to include & maintain such code in our library and/or if it's worth that.

dgladkov:
A lot of answers suggested in that SO post are impractical (writing manual SSE2 instructions, depending on lots of third party libraries), but the general idea is that the only way to gain meaningful performance improvement is to write native code. The best solutions for Python are [cffi](http://cffi.readthedocs.io/en/latest/) and [Cython](http://cython.org/). cffi is a better solution for small functions, while Cython is better when you need complex C<->Python interop, so I picked cffi for my test. My current C implementation looks like this (I know only basic C so this might be suboptimal):
```c
void xor_one(const char *data, const unsigned char key, char *result)
{
    unsigned long data_len = strlen(data);
    unsigned long idx;
[...]

PR #18, opened 2018-01-18
KaitaiStream code cleanup

arekbulski:
Those are just to make code readable/pythonic. I dont know how to run the benchmarks on this code so no performance guarantees, although it should be comparable.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -25,15 +25,13 @@ class KaitaiStruct(object):
     def __exit__(self, *args, **kwargs):
         self.close()
 
+    def close(self):
+        self._io.close()
+
     @classmethod
     def from_file(cls, filename):
-        f = open(filename, 'rb')
-        try:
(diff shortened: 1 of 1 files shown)


GreyCat:
I'm slightly concerned about performance of these `process_*` routines. @dgladkov did some benchmarking and suggested current solution as optimal. I believe you've started some work on Python benchmarks, so may be we should return to it after you'll be happy with the results?

arekbulski:
The difference is small, it does not warrant complexity of current implementation. I am familiar with his changes, read #8 . If you really want to speed it up, we need Cython for that. Those are just superficial speedups. 

I know where the benchmarks repo is but I dont know how to run it against a PR. I would require some assistance with that.

arekbulski:
I suggest we merge it now, and return to processxor performance when there is nothing better to do.

arekbulski:
Forgive me for merging my own PR, we can get back to this topic when Cython gets added.

PR #19, opened 2018-01-19
read_bytes min amount is 0

arekbulski:
Additional check when reading bytes, requesting negative amount should cause error. Construct has this check too.


GreyCat:
Makes sense, thanks! I thought that Python's stream `read` would have this check, but instead it works something like negative array indexes or something like that.

Issue #21, opened 2018-03-06
Alternative runtime without `struct`


GreyCat:
Python's `struct` module seems to be pretty inefficient for our purposes. Namely, in all APIs it provides, it requires passing a format string into `unpack`-like function, which then parses that format string in runtime, calls relevant unpack methods, and then constructs a tuple with a single value, which we extract right away.

Actually, `struct` even has everything we need — for example, [these are functions which read ("unpack") integers](https://github.com/python/cpython/blob/a5552f023e1d8cbafee1e51d316cc581deb2295f/Modules/_struct.c#L379-L475), but it's not exposed as Python API.

Would it make sense / be faster to introduce alternative, native Kaitai Struct API which would be written in C, but would be faster than existing one?

Cc @koczkatamas @KOLANICH @arekbulski

arekbulski:
I will reach out to python mailing list, the guys there are very helpful with advice, and knowledgable too.

arekbulski:
I will reach out to python mailing list, the guys there are very helpful with advice, and knowledgable too.

There is a way to pre-compile a formatstring into a packer object, but it does also return a tuple. 
https://docs.python.org/3/library/struct.html#classes

KOLANICH:
>Would it make sense / be faster to introduce alternative, native Kaitai Struct API which would be written in C

IMHO no.

1 C is fast, but I wonder if Rust may be better here.
2 If have a better API, let it be a part of python, not standalone library.
3 as @arekbulski has mentioned, it is possible to precompile the structs parsers. Of course it is the work of KSC to merge adjacent fields into a single struct if it is possible. I wonder if makes any sense to do some inlining.

```yaml
meta:
 id: LLL
seq:
 - id: a
  type: u4
 - id: b
  type: u4
 - id: c
  type: aa
 - id: d
  type: f8
types:
  aa:
    seq:
      - id: b
        type: u1
```

now (simplified, only conducts sense)
```python
class LLL:
  ...
  a=unpack("I", ...)
  b=unpack("I", ...)
  c=Aa(...)
  d=unpack("d")
```

with precompilation
```python
class LLL:
  ab_unp=Struct("II")
  d_unp=Struct("d")
  ...
  a, b=__class__.ab_unp.unpack(...)
  c=Aa(...)
  d = __class__.ab_unp.unpack(...)
```

with inlining:
```python
class LLL:
  abcd_unp=Struct("IIBd")
  ...
[...]

PR #22, opened 2018-03-12
replaced struct.unpack with struct.Struct().unpack

arekbulski:
Per issue #21 . I ran the tests, but leave the benchmarks to you. I am having problems with benchmarks/build-formats.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -1,5 +1,6 @@
 import itertools
 import sys
+import struct
 from struct import unpack
 from io import BytesIO  # noqa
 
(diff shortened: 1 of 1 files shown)


GreyCat:
Thanks! I've ran the simple PCAP-based benchmark.

* Before: 4.256415247915 ± 0.08523450167324
* After: 4.16146477063333 ± 0.047591945923799

The overall effect is ~2%, and is comparable to the margin of error (which is also about 1-2%). But 2% is still good, I guess ;)

kaitaistruct.py
```diff
@@ -1,5 +1,6 @@
 import itertools
 import sys
+import struct
 from struct import unpack
```
GreyCat: We can probably remove this line now.

arekbulski:
I made same upgrade to Construct, and the speedup was also within margin of error.

PR #29, opened 2018-04-06
relative seek(-1,CUR) instead of seek(tell()-1)

arekbulski:
Discussed in https://github.com/kaitai-io/kaitai_struct/issues/411.


GreyCat:
Cool, thanks!

PR #30, opened 2018-04-06
size() is precomputed for memory-based streams

arekbulski:
Discussed in https://github.com/kaitai-io/kaitai_struct/issues/411.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -50,6 +50,8 @@ class KaitaiStruct(object):
 class KaitaiStream(object):
     def __init__(self, io):
         self._io = io
+        if isinstance(io, BytesIO):
+            self.size_cached = self.size()
         self.align_to_byte()
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


GreyCat:
My point here is why we should *not* do things like this is that files are actually not constant size. There are live binary logs which keeps being appended to (as you run the parse on it), there are live network packet captures, etc.

In some cases, setting a single size is unavoidable (for example, memory mapping works this way: you have to know a size apriori), but for general case we don't want to cache it. If we still want to do it, then it should be behind some kind of opt-in mechanism (for example, a special option in a KaitaiStream constructor to control this size caching).

GreyCat:
If that's for BytesIO only:

* I believe you can do something like `_io.getbuffer().nbytes` in `size()`, and that could be marginally faster
* If that's slow / not available somehow, we can actually supply the size as argument

Or, better yet, let's just wait until we get real substreams running. They are bound to have size supplied to them.

arekbulski:
Okie dokie, I agree.

PR #39, opened 2020-02-03
Fixed setup.py and setup.cfg

KOLANICH:


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = kaitaistruct
-version = attr: kaitaistruct.__version__
+version = attr: __version__
```
dgelessus: What's the reason for this change? I haven't had any issues with the previous setting.

setup.py
```diff
@@ -1,10 +1,5 @@
...
+from kaitaistruct import __version__
 
-this_dir = os.path.dirname(__file__)
-cfg = read_configuration(os.path.join(this_dir, 'setup.cfg'))
-cfg["options"].update(cfg["metadata"])
-cfg = cfg["options"]
-
-setup(**cfg)
+if __name__ == "__main__":
+    setup()
```
dgelessus: Not sure if this guard is necessary - there's no legitimate use case for importing setup.py as a module. If anything, I would throw an exception if `__name__ != "__main__"`.

setup.cfg
KOLANICH: I am planning to not to import, but extract this var with `ast.parse`. I remember I have done it for some projects, but I don't remember their names, probably I would have to recode that from scratch.

setup.py
KOLANICH: The guard is not really necessary, but it is the right thing to add it, so I have done it.

>If anything, I would throw an exception if __name__ != "__main__".

IMHO we shouldn't. We cannot predict all the ways third party code can use our code and we shouldn't counteract it.

setup.cfg
dgelessus: Hm, not sure if this is worth the effort in our case. There are situations where `attr: module.__version__` doesn't work (see pypa/setuptools#1724, pypa/setuptools#1960), but `kaitaistruct` isn't affected by that.

If you really want to get `kaitaistruct.__version__` without importing the module, I would use the code from pypa/setuptools#1960 instead of reimplementing it from scratch.

setup.py
dgelessus: > We cannot predict all the ways third party code can use our code

Actually in this case we can - all previous versions of `kaitaistruct` did not have this guard, meaning `import setup` has never worked before, so nothing will be broken by throwing an exception in that case.

> and we shouldn't counteract it.

Yes, we should counteract ways to call our code that have no use and will only happen by accident.

setup.py
```diff
@@ -1,10 +1,29 @@
...
+except:
+    from pathlib2 import Path
 
-this_dir = os.path.dirname(__file__)
-cfg = read_configuration(os.path.join(this_dir, 'setup.cfg'))
-cfg["options"].update(cfg["metadata"])
-cfg = cfg["options"]
 
-setup(**cfg)
+def extract_global_vars(file):
```
dgelessus: I think it's a bit overkill to add a function for reading *all* global variables - realistically we only need `__version__`.

setup.py
```diff
@@ -1,10 +1,29 @@
...
+    t = file.read_text(encoding="utf-8")
+    m = ast.parse(t)
```
dgelessus: I would read the file in binary mode instead of hardcoding the encoding here. `ast.parse` accepts `bytes` input, in which case it lets the parser determine the encoding (this also respects `# coding: ...` comments and such).

In general I'm in favor of "UTF-8 everywhere", but in this case we specifically want to behave like Python's normal parser, so we shouldn't override its defaults.

setup.py
```diff
@@ -1,10 +1,29 @@
...
+    res = {}
+    for s in m.body:
+        if isinstance(s, ast.Assign):
+            vo = s.value
+            v = None
+            if isinstance(vo, ast.Str):
+                v = vo.s
+            elif isinstance(vo, ast.Num):
+                v = vo.n
```
dgelessus: You can use `ast.literal_eval` to safely evaluate a constant expression AST.

setup.py
```diff
@@ -1,10 +1,29 @@
...
+            
+            if v is not None:
+                for t in s.targets:
+                    if isinstance(t, ast.Name):
+                        res[t.id] = v
```
dgelessus: This isn't how assignment to multiple variables works. `x, y = 1, 2` sets `x = 1` and `y = 2`, but your code interprets it as `x = (1, 2)`, `y = (1, 2)`.

For simplicity I would ignore complex assignments with multiple targets, unpacking, etc. - realistically, `__version__` will only be set with a simple single-target assignment.

setup.py
```diff
@@ -1,10 +1,29 @@
...
+    return res
+
+if __name__ == "__main__":
+    setup(version=extract_global_vars(Path(__file__).absolute().parent / "kaitaistruct.py")["__version__"])
```
dgelessus: The `.absolute()` call shouldn't be necessary here - the code will work fine even if `__file__` is a relative path.

GreyCat:
Can you clarify what is the purpose of this PR? There's no description, and title is as ambiguous as it can get:

> Fixed setup.py and setup.cfg

What is/was wrong with setup.py/setup.cfg so it required fixing?

setup.py
KOLANICH: it is jecessary, I had problems with other packages. Order matters.

setup.py
```diff
@@ -1,10 +1,29 @@
-import os
 from setuptools import setup
-from setuptools.config import read_configuration
+import ast
+try:
+    from pathlib import Path
+except:
+    from pathlib2 import Path
```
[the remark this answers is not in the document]
KOLANICH: I would like to drop building on python 2 completely, since it is EOL, rather than use `os.path`. But if someone needs it, IMHO making him only to install `pathlib2` is less cruel than making him to rewrite half of the code.

setup.py
KOLANICH: Yes, it is a bit overkill. But who knows if we want to extract something else in future.

setup.py
KOLANICH: you are wrong. https://ideone.com/xashTQ

KOLANICH:
>What is/was wrong with setup.py/setup.cfg so it required fixing?

1. Obsolete approaches were used. Explicitly reading `setup.cfg` was no longer needed, it is read and merged automatically even if something is set in `setup()`. 
2. `pep517` is a uniform way to trigger building. Implemented its support.
3. extraction of version without importing. Useful when cwd is an another dir.

Also I wonder if it is acceptible to make `kaitaistruct` a dir, so we can capture the version from a git tag and save it into `version.py` using `setuptools_scm`.

setup.py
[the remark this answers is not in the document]
dgelessus: > I would like to drop building on python 2 completely

I thought we decided against that in kaitai-io/kaitai_struct#695?

> But if someone needs it, IMHO making him only to install `pathlib2` is less cruel than making him to rewrite half of the code.

Transitioning from Python 2 to 3 is besides the point here - we still support Python 2. The question is: when running on Python 2, what should our setup.py do:

* Use `pathlib2`, requiring Python 2 users to *manually* install an additional dependency, for minimally cleaner file manipulation code
* Use `open` and `os.path`, requiring no additional dependencies for users, at the cost of having to slightly rewrite two lines

I am very strongly for the second option. Requiring extra dependencies *to run setup.py* is very unusual, and I don't think it makes sense to break automatic package installation for all Python 2 users just so we have *slightly* cleaner code in our setup.py.

setup.py
dgelessus: Ah, you're right, my bad. In that case it actually makes sense to handle it this way (although IMHO it's still kind of unlikely that `__version__` would be set in a multi-target assignment).

setup.py
dgelessus: How would `.absolute()` make a difference though? If the input path is relative, `pathlib` will correctly return a new path (relative to the cwd) that points to kaitaistruct.py.

```python
>>> pathlib.Path("setup.py").parent / "kaitaistruct.py"                                         
PosixPath('kaitaistruct.py')
>>> pathlib.Path("../setup.py").parent / "kaitaistruct.py"                                      
PosixPath('../kaitaistruct.py')
>>> pathlib.Path("some/sub/dir/setup.py").parent / "kaitaistruct.py"                            
PosixPath('some/sub/dir/kaitaistruct.py')
>>> # Same with os.path
>>> os.path.join(os.path.dirname("setup.py"), "kaitaistruct.py")                                
'kaitaistruct.py'
>>> os.path.join(os.path.dirname("../setup.py"), "kaitaistruct.py")                             
'../kaitaistruct.py'
>>> os.path.join(os.path.dirname("some/sub/dir/setup.py"), "kaitaistruct.py")                   
'some/sub/dir/kaitaistruct.py'
```

Of course we're assuming that `__file__` is in fact relative to the current cwd, but if it isn't, `.absolute()´ wouldn't work correctly either.

KOLANICH: I remember that if my cwd is the parent dir of the parent dir of a file, without `absolute()` everything works incorrectly. =

setup.py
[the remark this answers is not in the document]
KOLANICH: * `pathlib` neither requires additional dependencies for python 3 users
* python 2 is EOL by now, why should we use `os.path` only to allow users of a deprecated dead version run `setup.py` without additional dependencies? They can always install a wheel, btw, or manually copy the file to the location where they need it.

higherorderfunctor:
> What is/was wrong with setup.py/setup.cfg so it required fixing?

For me, this fixed being able to install with poetry.

```
poetry add git+https://github.com/kaitai-io/kaitai_struct_python_runtime.git

[ParseVersionError]
Unable to parse "attr: kaitaistruct.__version__".
```

I did have to rename the branch though since it is named the same as a file in the repo.

```
poetry add git+https://github.com/KOLANICH/kaitai_struct_python_runtime.git#setup.cfg

[CalledProcessError]
Command '['git', '--git-dir', '/tmp/pypoetry-git-kaitai_struct_python_runtime5m5xquv5/.git', '--work-tree', '/tmp/pypoetry-git-kaitai_struct_python_runtime5m5xquv5', 'checkout', 'setup.cfg']' returned non-zero exit s
tatus 128.
```

Clone of this PR with a different branch name.

```
poetry add git+https://github.com/higherorderfunctor/kaitai_struct_python_runtime.git#setup_fix

Updating dependencies
Resolving dependencies... (4.0s)

Writing lock file


Package operations: 1 install, 0 updates, 0 removals

  - Installing kaitaistruct (0.9 1d01b14)
```

KOLANICH:
>For me, this fixed being able to install with poetry.

please don't use `poetry`, it is destroying the ecosystem because it is harmful because it is wasteful, but matches well the current fashionable cargo cult :(

dgelessus:
@higherorderfunctor Although I don't agree with @KOLANICH's opinion of Poetry, if it fails to parse a setup.cfg with `version = attr: ...` in it, that's a bug in Poetry and not in the setup.cfg. (I'm actually surprised that Poetry accepts the new setup.cfg from this PR, considering that it still uses the `attr: ...` feature. Most likely it's interpreting the version as literally `attr: __version__`, as it looks like Poetry's [code for parsing setup.cfg](https://github.com/python-poetry/poetry/blob/96191ca4f7551243366bda030426995fce5c4b57/poetry/utils/setup_reader.py#L99) never handles `attr: ...` anywhere.)

> I did have to rename the branch though since it is named the same as a file in the repo.

That is also a bug in Poetry - it's perfectly legal to have a Git branch name that is identical to the name of a file in that branch.

I'd encourage you to report both of these bugs on the Poetry issue tracker: https://github.com/python-poetry/poetry/issues

higherorderfunctor:
> > I did have to rename the branch though since it is named the same as a file in the repo.
> 
> That is also a bug in Poetry - it's perfectly legal to have a Git branch name that is identical to the name of a file in that branch.

I believe you have to add -- to the checkout command to disambiguate file vs branch.  Poetry isn't doing this and as far as I know there is no way to inject this.

https://stackoverflow.com/a/25322508

dgelessus (changes requested):
Well, this PR now does something completely different than before... This isn't "fixing" setup.py and setup.cfg, it's replacing them entirely with a pyproject.toml-based configuration.

Unfortunately, at this point, this is a complete non-starter for us. According to [the setuptools documentation about `pyproject.toml` configuration](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html):

> ✏️ **Note**
> 
> New in 61.0.0 (**experimental**)

> ⚠️ **Warning**
> 
> Support for declaring [project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/) or configuring `setuptools` via `pyproject.toml` files is still experimental and might change (or be removed) in future releases.

We shouldn't rely on an experimental unstable feature that's [not even a month old](https://setuptools.pypa.io/en/latest/history.html#v61-0-0), unless there's a very good reason for it, and I'm not seeing that here - the current setup.py/setup.cfg work just fine.

KOLANICH:
>(or be removed) in future releases.

feels like bullshit. PEP 621 is accepted, had been long a goal for `setuptools` and now it is implemented, there is no way back. If you dislike relying on `setuptools` experimental feature (it contains the bug I have mentioned, that's why `version` is not `dynamic` now), I can add 5 lines (2 to enable `flit_core` as a build backend, 1 a section header, 1 specifying the source file for `flit` and 1 empty line) and comment out the 2 lines enabling `setuptools` as a build backend, and the stuff will be built with `flit_core`, the `project` section stays the same.

>Regardless of whether the feature is stable or not, requiring a new setuptools version is also a problem for Python version compatibility. setuptools 61.0.0 requires at least Python 3.7 (59.7.0 dropped support for older Python versions). Kaitai Struct still supports Python 3.4 and 2.7, so we can't require anything newer than setuptools 43.0.0 (from late 2019), the last version to support Python 3.4.

I wonder if we can backport the latest `setuptools` to 3.4. Usually python version requirement is raised because of often unjustified (to the point of it being a cargo cult) use of so called `f-strings` and typing hints for variables within function bodies. I guess it can be possible to write an AST transformer undoing it.

KOLANICH:
Also `requires-python` specifies the version of a package to run the code on, not the one to build a wheel on.

generalmimon:
@KOLANICH:

> If you dislike relying on `setuptools` experimental feature (it contains the bug I have mentioned, that's why `version` is not `dynamic` now), I can add 5 lines (...), and the stuff will be built with `flit_core`, the `project` section stays the same.

> I wonder if we can backport the latest `setuptools` to 3.4. (...) I guess it can be possible to write an AST transformer undoing it.

But _why_ - what would be the point? Why not just use something established and stable?

KOLANICH:
1. It looks more tidy. I'm a mad perfectionist, you know.
2. It is more standardized and has more tools around it. And will likely have even more tools thanks to network effects.
3. `setup.cfg` is planned to be dropped in indefinite future.

KOLANICH:
Anyway, I revert back to `setup.cfg`. The metadata can be moved into `pyproject.toml` in an another PR.

dgelessus (approved):
These changes all make sense IMO. Just one suggestion:

setup.cfg
```diff
@@ -3,9 +3,9 @@ name = kaitaistruct
 version = attr: kaitaistruct.__version__
 author = Kaitai Project
 author_email = greycat@kaitai.io
-url = http://kaitai.io
+url = https://kaitai.io
 description = Kaitai Struct declarative parser generator for binary data: runtime library for Python
-long_description = file: README.rst
+long_description = file: README.md
```
dgelessus, line 8: Just noticed that we also need to set `long_description_content_type`, otherwise it defaults to `text/x-rst`, which is not correct. (`twine check` warns about this.)

```suggestion
long_description = file: README.md
long_description_content_type = text/markdown
```

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 36.7.0", "wheel >= 0.23.0"]
```
dgelessus, line 2: `long_description_content_type` is only supported since setuptools 38.6.0.

```suggestion
requires = ["setuptools >= 38.6.0", "wheel >= 0.23.0"]
```

setup.cfg
```diff
@@ -31,16 +31,18 @@ zip_safe = True
 include_package_data = True
 py_modules = kaitaistruct
 python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*
+install_requires =
+    enum34; python_version < "3.4"
+setup_requires =
+    setuptools >= 36.7.0
```
dgelessus, line 37:
```suggestion
    setuptools >= 38.6.0
```

pyproject.toml
generalmimon: @KOLANICH This was _not_ resolved in [`bfacaec...5858f14`](https://github.com/kaitai-io/kaitai_struct_python_runtime/compare/bfacaec50a0386ffa3d768ec701a3ca923f41068..5858f1401a8c3581b91e0daff4367c7d66604507) - you relaxed the `setuptools` version requirement to <code>&gt;= 3<b><ins>6</ins></b>.6.0</code> but it should have been **increased** to <code>&gt;= 3<b><ins>8</ins></b>.6.0</code>.

KOLANICH:
>Why don't you just use the Commit suggestion button that GitHub offers?

Most of JS-based functionality of GitHub is broken for me by now.

dgelessus (approved):
Thanks!

PR #42, opened 2020-03-16
Simplify bytes_strip_right and bytes_terminate implementations

dgelessus:
Python has builtins that do almost exactly what these methods do.

This is probably also much faster than before, because the builtins are implemented in C rather than pure Python, but I haven't tested/measured this.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -444,6 +444,7 @@ class ValidationGreaterThanError(ValidationFailedError):
         self.max = max
         self.actual = actual
 
+
 class ValidationNotAnyOfError(ValidationFailedError):
     """Signals validation failure: we required "actual" value to be
     from the list, but it turned out that it's not.
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -306,31 +306,14 @@ class KaitaiStream(object):
 
     @staticmethod
     def bytes_strip_right(data, pad_byte):
-        new_len = len(data)
-        if PY2:
-            # data[...] must yield an integer, to compare with integer pad_byte
-            data = bytearray(data)
-
-        while new_len > 0 and data[new_len - 1] == pad_byte:
-            new_len -= 1
-
-        return data[:new_len]
+        return data.rstrip(KaitaiStream.byte_from_int(pad_byte))
 
     @staticmethod
     def bytes_terminate(data, term, include_term):
-        new_len = 0
-        max_len = len(data)
-        if PY2:
-            # data[...] must yield an integer, to compare with integer term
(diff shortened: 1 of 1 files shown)


dgelessus:
(the code style violation isn't actually from my change, but I'm fixing it here anyway because I don't want to make a second PR just for that)

generalmimon:
> This is probably also much faster than before, because the builtins are implemented in C rather than pure Python, but I haven't tested/measured this.

I have, and here is what I got:

**`bytes_strip_right`**
| Branch ╲ Python | 2.7 | 3.8 | pypy2 | pypy3 |
| - | - | - | - | - |
| [`kaitai-io:master`](https://github.com/kaitai-io/kaitai_struct_python_runtime/tree/master) | 10.819 | 300.859 | 5.069 | 4.121 |
| [`dgelessus:simplify_terminate_strip_right`](https://github.com/dgelessus/kaitai_struct_python_runtime/tree/simplify_terminate_strip_right) | 0.001 | 7.059 | 0.004 | 1.973 |

**`bytes_terminate`**
| Branch ╲ Python | 2.7 | 3.8 | pypy2 | pypy3 |
| - | - | - | - | - |
| [`kaitai-io:master`](https://github.com/kaitai-io/kaitai_struct_python_runtime/tree/master) | 976.8 | 228.9 | 29.6 | 4.1 |
| [`dgelessus:simplify_terminate_strip_right`](https://github.com/dgelessus/kaitai_struct_python_runtime/tree/simplify_terminate_strip_right) | 9.0 | 1.9 | 33.9 | 1.4 |

_(all numbers are rounded median execution times in milliseconds from ~1000 executions of the particular method - see [`test.py`](https://github.com/generalmimon/ks-python-runtime-setuptools-test/blob/d4087af/test.py) and [CI results](https://github.com/generalmimon/ks-python-runtime-setuptools-test/actions/runs/307057129) for more info)_

As you said, the only explanation I have is the new implementation is written in C instead of Python, but still, I wouldn't have thought that the differences would be so striking ;-)

generalmimon (approved):
A very good improvement, thanks!

dgelessus:
Thank you for the measurements! Yeah, CPython really isn't very fast when it comes to simple data manipulation... I'm a bit surprised that there's a noticeable difference even with PyPy - clearly the JIT already helped a lot with the performance before, but it's even faster with the built-in methods. (Well, except for that one case with `bytes_terminate` on PyPy 2, which somehow became a bit *slower*. Not quite sure what's going on there - I'm guessing that the way I wrote the Python 2/3 compatibility code confuses PyPy 2 and prevents optimization somehow...)

PR #48, opened 2020-08-03
Add methods for serialization

UserUnknownFactor:
It would be nice to have the python library prepared for serialization functionality.

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -119,6 +119,9 @@ class KaitaiStream(object):
     packer_u8le = struct.Struct('<Q')
 
     # ------------------------------------------------------------------------
+    # Reading
+    # ------------------------------------------------------------------------
+    # ------------------------------------------------------------------------
     # Signed
     # ------------------------------------------------------------------------
 
@@ -213,6 +216,95 @@ class KaitaiStream(object):
     def read_f8le(self):
         return KaitaiStream.packer_f8le.unpack(self.read_bytes(8))[0]
 
+    # ------------------------------------------------------------------------
+    # Writing
+    # ------------------------------------------------------------------------
+    # ------------------------------------------------------------------------
+    # Signed
+    # ------------------------------------------------------------------------
+
+    def write_s1(self, data):
+        return self.write_bytes(KaitaiStream.packer_s1.pack(data))
+
+    # ........................................................................
+    # Big-endian
+    # ........................................................................
+
+    def write_s2be(self, data):
+        return self.write_bytes(KaitaiStream.packer_s2be.pack(data))
+
+    def write_s4be(self, data):
+        return self.write_bytes(KaitaiStream.packer_s4be.pack(data))
+
+    def write_s8be(self, data):
+        return self.write_bytes(KaitaiStream.packer_s8be.pack(data))
+
+    # ........................................................................
+    # Little-endian
+    # ........................................................................
+
+    def write_s2le(self, data):
+        return self.write_bytes(KaitaiStream.packer_s2le.pack(data))
+
+    def write_s4le(self, data):
+        return self.write_bytes(KaitaiStream.packer_s4le.pack(data))
+
+    def write_s8le(self, data):
+        return self.write_bytes(KaitaiStream.packer_s8le.pack(data))
+
+    # ------------------------------------------------------------------------
+    # Unsigned
+    # ------------------------------------------------------------------------
+
+    def write_u1(self, data):
+        return self.write_bytes(KaitaiStream.packer_u1.pack(data))
+
+    # ........................................................................
+    # Big-endian
+    # ........................................................................
+
+    def write_u2be(self, data):
+        return self.write_bytes(KaitaiStream.packer_u2be.pack(data))
+
+    def write_u4be(self, data):
+        return self.write_bytes(KaitaiStream.packer_u4be.pack(data))
+
+    def write_u8be(self, data):
+        return self.write_bytes(KaitaiStream.packer_u8be.pack(data))
+
+    # ........................................................................
+    # Little-endian
+    # ........................................................................
+
+    def write_u2le(self, data):
+        return self.write_bytes(KaitaiStream.packer_u2le.pack(data))
+
+    def write_u4le(self, data):
+        return self.write_bytes(KaitaiStream.packer_u4le.pack(data))
+
+    def write_u8le(self, data):
+        return self.write_bytes(KaitaiStream.packer_u8le.pack(data))
+
+    # ........................................................................
+    # Big-endian
+    # ........................................................................
+
+    def write_f4be(self, data):
+        return self.write_bytes(KaitaiStream.packer_f4be.pack(data))
+
+    def write_f8be(self, data):
+        return self.write_bytes(KaitaiStream.packer_f8be.pack(data))
+
+    # ........................................................................
+    # Little-endian
+    # ........................................................................
+
+    def write_f4le(self, data):
+        return self.write_bytes(KaitaiStream.packer_f4le.pack(data))
+
+    def write_f8le(self, data):
+        return self.write_bytes(KaitaiStream.packer_f8le.pack(data))
+
     # ========================================================================
     # Unaligned bit values
     # ========================================================================
@@ -293,6 +385,48 @@ class KaitaiStream(object):
             )
         return r
 
+    def write_bytes(self, data):
+        if data is None:
+            return
+        nb = len(data)
+        if nb > 0 and self._io.write(data) != nb:
+            raise Exception(
+                "not all of %d bytes written" %
+                (nb,)
+            )
+        return nb
+
+    def write_terminator(self, term_byte):
+        return self._io.write(KaitaiStream.packer_u1.pack(term_byte))
+
+    def write_padding(self, actual_size, size, pad_byte):
+        pad = size - actual_size
+        if pad > 0:
+            if pad_byte is not None:
+                return self._io.write(KaitaiStream.packer_u1.pack(pad_byte) * pad)
+            else:
+                raise Exception("no padding filler provided")
+        return 0
+
+    def write_bytes_term(self, data, size, term_byte, pad_byte):
+        if data is None or not size:
+            return
+        nb = len(data)
+        if nb < size:
+            if nb > 0 and self._io.write(data) != nb:
+                raise Exception(
+                    "not all of %d bytes written" %
+                    (nb,)
+                )
+        else:
+            raise Exception(
+                "Writing %d bytes, but %d bytes (including terminator) were given" %
+                (size, nb + 1)
+            )
+        self.write_terminator(term_byte)
+        self.write_padding(nb, size - 1, pad_byte)
+        return size
+
     def read_bytes_full(self):
         return self._io.read()
 
@@ -302,7 +436,7 @@ class KaitaiStream(object):
             c = self._io.read(1)
             if c == b'':
                 if eos_error:
-                    raise Exception(
+                    raise EOFError(
                         "end of stream reached, but no terminator %d found" %
                         (term,)
                     )


KOLANICH:
#13 and #12 are related

kaitaistruct.py
```diff
@@ -1,18 +1,19 @@
 import itertools
 import sys
 import struct
+import zlib
```
KOLANICH, line 4: No, it shouldn't be here. Instead https://github.com/kaitai-io/kaitai_compress/pull/2 (and maybe its successor https://github.com/KOLANICH/transformerz.py/tree/master/transformerz/kaitai ) should be used ... in a generic way. In fact it really requires the interfaces proposal being implemented. So, no compression in serialization for now.

kaitaistruct.py
```diff
@@ -1,18 +1,19 @@
...
 from io import open, BytesIO, SEEK_CUR, SEEK_END  # noqa
 
-PY2 = sys.version_info[0] == 2
+PY_OLD = sys.version_info[0] == 2 or (sys.version_info[0] == 3 and sys.version_info[1] < 2)
 
-# Kaitai Struct runtime streaming API version, defined as per PEP-0396
-# standard. Used for two purposes:
+# Kaitai Struct runtime streaming API version, defined as per PEP-0396 standard.
```
KOLANICH, line 9: Please no formatting-only changes. If you want to bring them, they must be in a separate PR with bulletproof reasoning why they are needed.

kaitaistruct.py
```diff
@@ -30,22 +31,30 @@ def close(self):
 
     @classmethod
     def from_file(cls, filename):
-        f = open(filename, 'rb')
-        try:
+        with open(filename, 'rb') as f:
```
KOLANICH, line 34: You may also wanna look at #25

kaitaistruct.py
UserUnknownFactor: Done. I've reverted them.

kaitaistruct.py
UserUnknownFactor: I see that's a thing to consider. Thank you.

kaitaistruct.py
UserUnknownFactor: Please, for the love of god add serialization option to either of those, then and ship them with this runtime module. It feels very clumsy to have to write the code for those from scratch each time or install different module for the same single purpose when the compression/decompression is clearly supported by the ksy format natively.

UserUnknownFactor:
I've fixed the stuff.

kaitaistruct.py
KOLANICH: I don't understand what you mean. My modules define interface of 3 functions for compressor/decompressors and implement them. And they are actually used in the wild. I use my module to compress a database.

kaitaistruct.py
```diff
@@ -279,57 +351,67 @@ def read_bits_int_le(self, n):
     # Byte arrays
     # ========================================================================
 
-    def read_bytes(self, n):
+    def alignment(self, a):
```
KOLANICH, line 354: needs docs

kaitaistruct.py
UserUnknownFactor: I mean wouldn't it be better if I could just run `pip install kaitaistruct` and use the decoder/encoder right away?

kaitaistruct.py
UserUnknownFactor: What format and where should I put it?

kaitaistruct.py
KOLANICH: No. This stuff is optional.

kaitaistruct.py
KOLANICH: It is currently not very what it does and how to use it and why it is implemented this way and not another one.

kaitaistruct.py
```diff
@@ -279,57 +351,67 @@ def read_bits_int_le(self, n):
...
+        return (a - self.pos()) % a
+
+    def read_bytes(self, n, align=0):
```
dgelessus, line 357: I see that you're adding new arguments to a few existing functions. In terms of backwards compatibility that's okay, as long as you provide appropriate defaults (which you do, so that's not an issue). But it would be useful to have some context regarding the added arguments - what are they needed for, and how are they meant to be used in code generated by KSC?

About the `align` argument in particular, I assume that is meant to allow reading data that is aligned to a n-byte boundary in the data? The KSY language doesn't have native support for this, so I don't understand how this parameter is meant to be used. If you're interested in implementing native support for alignment, that would be better done in a separate PR (see also the previous discussion about this kind of feature in kaitai-io/kaitai_struct#12).

kaitaistruct.py
```diff
@@ -279,57 +351,67 @@ def read_bits_int_le(self, n):
...
+            self._io.write(padding * pad)
+        if align > 1:
+            self._io.write(padding * self.alignment(align))
+        return
+
     def read_bytes_full(self):
         return self._io.read()
 
-    def read_bytes_term(self, term, include_term, consume_term, eos_error):
+    def read_bytes_term(self, term, include_term=False, consume_term=True, eos_error=True, elem_size=1):
```
dgelessus, line 384: As above, what's the purpose of the `elem_size` parameter, and how exactly is it related to serialization?

Also, I'm not sure if it's very useful to add default values for existing arguments. All code that uses this API is generated by the compiler, which already sets all these arguments explicitly, so these defaults would never be used.

kaitaistruct.py
```diff
@@ -393,10 +475,10 @@ def process_rotate_left(data, amount, group_size):
     # ========================================================================
 
     @staticmethod
-    def int_from_byte(v):
-        if PY2:
+    def int_from_byte(v, signed=False):
+        if PY_OLD:
             return ord(v)
-        return v
+        return int.from_bytes(v, signed=signed)
```
dgelessus, line 481: `int_from_byte` is only a Python 2/3 compatibility method, for converting a 1-byte string into an integer byte value (needed because of API differences between Python 2 `str` and Python 3 `bytes`). I don't think it makes sense to add support for reading signed integers here.

In this state the implementation also doesn't work correctly on Python 2, because the `signed` paramter is ignored in the `PY2`/`PY_OLD` branch.

kaitaistruct.py
```diff
@@ -38,14 +38,32 @@ def from_file(cls, filename):
             f.close()
             raise
 
+    @classmethod
+    def to_file(cls, filename, data=None):
+        f = open(filename, 'rb')
+        try:
+            return cls(KaitaiStream(f), _mode='w', _data=data)
```
dgelessus, line 45: What are the `_mode` and `_data` parameters supposed to do? They aren't accepted by `KaitaiStruct.__init__` or by the `__init__`s generated by the compiler.

kaitaistruct.py
UserUnknownFactor: It's addressing this: https://github.com/kaitai-io/kaitai_struct/issues/12#issuecomment-277757932 Some structures need alignment to be read (like Unity serialized files for example). It's convenient to have this ready for such cases and maybe for expanding ksc format with this at some point.

kaitaistruct.py
```diff
@@ -279,57 +351,67 @@ def read_bits_int_le(self, n):
...
-    def read_bytes(self, n):
+    def alignment(self, a):
+        return (a - self.pos()) % a
+
+    def read_bytes(self, n, align=0):
         if n < 0:
-            raise ValueError(
-                "requested invalid %d amount of bytes" %
-                (n,)
-            )
```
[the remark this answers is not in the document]
UserUnknownFactor: Ok.

kaitaistruct.py
UserUnknownFactor: Why not? We can use it as a service method to convert integers this way. Also 3.2 version check was added exactly for `from_bytes` method as it is not implemented prior.

kaitaistruct.py
UserUnknownFactor: `__init__`s generated by the compiler (or manually currently)

kaitaistruct.py
dgelessus: When targetting Python, the compiler currently converts `process: zlib` into a `zlib.decompress` call directly in the generated code. As far as I can tell, the same approach would work with serialization, so there should be no need for extra functions in the runtime for this.

@KOLANICH This really isn't the place to propose major changes to how the standard KS `process` functions should be handled. At the moment, the compiler has native support for `process: zlib` when parsing, so it should be the same for serialization. If you want to change that, please discuss that in another issue.

kaitaistruct.py
UserUnknownFactor: Users can change code generated by compiller because it can't cover all use cases, in this way the code may be called manually. It's a chore to list all parameters in this case.

kaitaistruct.py
UserUnknownFactor: But `zlib` mentioned in the `process` options on the main kaitai.io site, how can it be optional?

kaitaistruct.py
```diff
@@ -38,14 +38,32 @@ def from_file(cls, filename):
...
+            f.close()
+            raise
+
     @classmethod
     def from_bytes(cls, buf):
         return cls(KaitaiStream(BytesIO(buf)))
 
+    @classmethod
+    def to_bytes(cls, buf, data=None):
+        return cls(KaitaiStream(BytesIO(buf)), _mode='w', _data=data)
```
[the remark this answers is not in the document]
UserUnknownFactor: The `_io` attribute of KaitaiStruct can be used for accessing the stream. And `buf` used for populating the stream with previous data, am I wrong?

kaitaistruct.py
UserUnknownFactor: Yes, the code is from that issue. I want to implement it but I am not sure I'll have time and it's useful to have as a service method anyway as there are many use-cases in the wild. That's why that thread is so active.

kaitaistruct.py
dgelessus: This method is meant to be used in contexts like this:

```python
byte_string = b"hello" # str in Python 2, bytes in Python 3
print(KaitaiStream.int_from_byte(byte_string[2]))
for byte in byte_string:
    print(KaitaiStream.int_from_byte(byte))
```

In Python 2, byte strings (`str`) are sequences of 1-byte strings. In Python 3, byte strings (`bytes`) are sequences of integers. `int_from_bytes` is used to get a consistent behavior on both versions, by converting Python 2's 1-byte strings into integers, and returning the argument unchanged on Python 3 (because it's already an integer).

If you need a way to read signed 1-byte integers, you can use the already existing `packer_s1` struct object.

kaitaistruct.py
UserUnknownFactor: What I meant when adding that is that compiller should probably reference runtime, not write code directly each time like it does now. But this isn't really place to discuss this, after some deliberation. 

If compiler don't reference it why do we need runtime in the first place, because compiler can just write `struct`s with all needed struct-format and service functions directly into resulting file which will be arguably more convenient to use as format classes would be self-contained then.

kaitaistruct.py
UserUnknownFactor: Here is use-case of what it could look like, when implemented on the compiler side...

Generated class:
```python
class UnityStrDat(KaitaiStruct):
    def __init__(self, _io, _parent=None, _root=None, _mode="r", _data=None):
        self._io = _io
        self._parent = _parent
        self._root = _root if _root else self
        if _mode == "r": self._read()
        elif _mode == "w" and _data: self._write(_data)

    def _read(self):
        self.header = UnityStrDat.HeaderRaw(self._io, self, self._root)
        self.num_strings = self._io.read_u4le()
        self.strings_arr = [None] * (self.num_strings)
        for i in range(self.num_strings):
            self.strings_arr[i] = UnityStrDat.StringStruct(self._io, self, self._root)

    def _write(self, data):
        if not data: return
        UnityStrDat.HeaderRaw(self._io, self, self._root, _mode="w", _data=data[0])
        self._io.write_u4le(len(data[1]) - 1)
        for i in range(len(data[1])):
            UnityStrDat.StringStruct(self._io, self, self._root, _mode="w", _data=data[1][i][1])


    class HeaderRaw(KaitaiStruct):
        def __init__(self, _io, _parent=None, _root=None, _mode="r", _data=None):
            self._io = _io
            self._parent = _parent
            self._root = _root if _root else self
            if _mode == "r": self._read()
            elif _mode == "w" and _data: self._write(_data)

        def _read(self):
            self.header_raw_bytes = self._io.read_bytes(500)

        def _write(self, data=None):
            if not data and self.header_raw_bytes: data = self.header_raw_bytes
            l = len(data)
            if not data or 500 != l:
                raise Exception("No header provided or wrong size (%d, instead of %d)" % (l, self._length))
            self._io.write_bytes(data)

    class StringStruct(KaitaiStruct):
        def __init__(self, _io, _parent=None, _root=None, _mode="r", _data=None):
            self._io = _io
            self._parent = _parent
            self._root = _root if _root else self
            if _mode == "r": self._read()
            elif _mode == "w" and _data: self._write(_data)

        def _read(self):
            self.s_size = self._io.read_u4le()
            self.string = self._io.read_bytes(self.s_size, align=4).decode(u"utf-8")

        def _write(self, data):
            if not data: return
            data = data.encode(self._encoding)
            self._io.write_u4le(len(data))
            self._io.write_bytes(data, align=4)
```
With usage like this:
```python

    if app_args.pack:
        a = UnityStrDat.from_file(file_name)
        a.close()
        data = load_old(os.path.splitext(file_name)[0]+'_data.old')
        new_name = "os.path.splitext(file_name)[0] + "_fixed" + os.path.splitext(file_name)[1]
        path_out = os.path.dirname(new_name)
        if path_out != '' and not os.path.exists(path_out):
            os.makedirs(path_out, exist_ok=True)
        a = UnityStrDat.to_file(new_name, (a.header1.header_raw_bytes, data))
        a.close()
    else:
        data = []
        with UnityStrDat.from_file(file_name) as a:
            for i in a.strings_arr:
                data.append([i.string, ''])

File: kaitaistruct.py (hunk 7 above)
Comment (UserUnknownFactor) [reply kaitaistruct.py]: Sorry, it was really dumb of me. :facepalm:

File: kaitaistruct.py (hunk 6 above)
Comment (UserUnknownFactor) [reply kaitaistruct.py:390]: For example `elem_size` can be used to deal with 2 or 4 byte strings.

File: kaitaistruct.py (hunk 5 above)
Comment (UserUnknownFactor) [reply kaitaistruct.py:357]: I've written example code below. Should I really make the separate PR? Neither functionality is done on the compiler side right now and it would be a nice demonstration of how it could work so maybe someone, or myself when I have a free time, can pick on it and add them to the compiler.

**Comment by UserUnknownFactor** [comment]
Example of code: https://github.com/kaitai-io/kaitai_struct_python_runtime/pull/48#discussion_r464776107
Also do `KaitaiStruct`s runtime needs `__init__` mention the stuff?
Because currently  compiler creates code that doesn't call `super` anyway.

File: kaitaistruct.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (UserUnknownFactor) [reply kaitaistruct.py:57]: What would be the best considering the example below?
Leave it as is? Or maybe add getter like `def get_stream(self): return self._io`?

File: kaitaistruct.py (hunk 8 above)
Comment (KOLANICH) [reply kaitaistruct.py:45]: Separate modes for reading and writing is IMHO a bad idea. It would require reopening the format. In the serialization proposal it was proposed to not to serialize from scratch, but cache the changes in memory and then apply them. So, if we ever have modes, there will be only 2, read-only and read-write. If we have 2 modes, they can be toggled by a boolean variable.

Comment (UserUnknownFactor) [reply kaitaistruct.py]: If you can implement cached changes I am all for it.

There can be more modes possible as underlying streams can support more, like appending mode of file IO stream so it's good idea not to shoot ourselves in the leg and not to restrict modes to just boolean.

**Comment by UserUnknownFactor** [comment]
I removed all changes related to possible compiler implementation and alignment, I'll do them as a separate PRs.

File: kaitaistruct.py [hunk 11]
```diff
@@ -317,6 +386,9 @@ def read_bytes_term(self, term, include_term, consume_term, eos_error):
             else:
                 r += c
 
+    def write_bytes_term(self, data, term=b'\0'):
```
generalmimon, line 389: AFAIR, no such method is used by the compiler branch [`serialization`](https://github.com/kaitai-io/kaitai_struct_compiler/tree/serialization) currently (it writes the terminator byte manually using `io.writeU1()` in Java if I remember correctly, so it doesn't make sense to have it in the runtime library. Moreover, using `terminator` as padding is actually incorrect, because `terminator` is not the same as `pad-right`. Terminator is a single byte signaling the end of some particular data, whereas padding is a continuous sequence of the `pad-right` byte after the end of data to fill the space reserved for the field using `size` or `size-eos: true` in the `.ksy` spec.

The `terminator` and `pad-right` bytes are _usually_ the same, but they don't have to be, see [`bytes_pad_term. ksy`](https://github.com/kaitai-io/kaitai_struct_tests/blob/master/formats/bytes_pad_term.ksy).

UserUnknownFactor: There is currently **no** Python serialization implementation whatsoever. So this entire PR is more of a suggestion and, currently, manual implementation helper. There is no need to do exactly the same thing Java does because Python's a different language and serialization in different languages can differ so it needn't be copied from one another as for ease of maintaining together. 

The main idea of separate function for terminator is possibility of 2 and 4 byte string support where we can't actually write a single byte terminator. And if the compuller would write it all directly just what do we even need the runtime for.

Write padding should be done using `write_bytes` and termination with `write_bytes_term` for code readability purposes. So if the compiller will write `write_bytes_term` it'd be easier to understand that it's a terminated block and not some random byte written. So there is no contradiction between `terminator` and `pad-right`.

kaitaistruct.py
KOLANICH: >There can be more modes possible as underlying streams can support more, like appending mode of file IO stream so it's good idea not to shoot ourselves in the leg by restricting modes to just boolean.

OK, append-only mode is needed too, because of FS_APPEND_FL.

kaitaistruct.py
generalmimon: > So this entire PR is more of a suggestion and, currently, **manual implementation helper**.

Kaitai Struct runtime libraries are not primarily intended to be a _manual implementation helper_. There are probably tons of other Python libraries that would serve this purpose better, or you can write your own. I'm sorry, but I'll not accept this as an argument.

> There is no need to do exactly the same thing Java does because Python's a different language

I'd say the opposite: there **must** be a need why do another thing than the existing languages because of robustness and maintainability. Please prepare a bulletproof explanation preferably with a solid reproducible benchmark (if it's from some performance reasons) every time you suggest such a change.

A few months ago I wanted to add support for [bit-sized integers](http://doc.kaitai.io/user_guide.html#_bit_sized_integers) in little-endian byte order (only big-endian bit ints had been supported before), which I implemented by renaming the existing `read_bits_int()` method to `read_bits_int_be()` and adding a new method `read_bits_int_le()` as its counterpart (which is the same algorithm with 2 small changes). It went quite well, because most runtime libraries were using the exact same logic and approach, so I quickly saw any differences which could cause bugs and could fix them immediately. In fact, when thinking about the common algorithm for a while, I was able to figure out a tiny optimization, which I applied to the rest of runtimes. 

The `ReadBitsInt()` method in Go was the only one implemented in a **completely** different way. There was no sign of the common algorithm I was used to, and thus I had to spend several hours figuring out how it works so that I could add the `ReadBitsIntLe()` counterpart. In the end, I gave up and rewrote the method completely from scratch using the established algorithm, which took me only a few minutes (see https://github.com/kaitai-io/kaitai_struct_go_runtime/pull/25). 

It's not only about maintainability, but the uncommon `ReadBitsInt()` algorithm actually had some real bugs that nobody noticed for a long time (https://github.com/kaitai-io/kaitai_struct_go_runtime/pull/22). I'd say unsurprisingly, because only a small subset of people are interested in Go codegen, and only units of people probably understood that custom algorithm.

The reason that _Python is a different language of Java_ is completely out of the game. You know very well that you can implement the same algorithm in all programming languages using often only simple 1:1 conversion: if you assign to a variable `myVar` in Java, in Python you assign to a variable `my_var`, and that's the whole difference between them.

> The main idea of separate function for terminator is possibility of 2 and 4 byte string support where we can't actually write a single byte terminator.

No, don't think about adding serialization support for things we don't have even parsing support, without **any** previous analysis of the implementation approach you choose. First read https://github.com/kaitai-io/kaitai_struct/issues/158 and https://github.com/kaitai-io/kaitai_struct/issues/187, then suggest some solution, which can be further discussed.

> Write padding should be done using `write_bytes` and termination with `write_bytes_term` for code readability purposes. So if the compiler will write `write_bytes_term` it'd be easier to understand that it's a terminated block and not some random bytes written.

This doesn't seem to me like a solid argument to me. It's a matter of personal preference, one could say that it's more explicit and transparent when you write the terminator manually. There has to be some other reason why deviate from the established convention from Java.

kaitaistruct.py
```diff
@@ -327,7 +399,7 @@ def ensure_fixed_contents(self, expected):
         return actual
 
     @staticmethod
-    def bytes_strip_right(data, pad_byte):
+    def bytes_strip_right(data, pad_byte=b'\0'):
```
generalmimon, line 402: This default value is wrong: the integer value of the byte is expected, not the "byte" itself. You would know this if you had inspected either to the function body or the generated Python code by KSC. If someone would rely on this default, it would keep returning the empty byte array regardless of the `data` (because the condition `data[new_len - 1] == pad_byte` would be always `False`, as you're comparing `int` to `bytes`, and `0 == b'0'` is `False` in both Python 2 and 3). 

Moreover, I also don't really see the need for adding the default values. As I said, KS runtime libraries are not meant to be _manual implementation helpers_.

And I can't recommend making that `pad_byte` parameter optional in this particular case anyway - it's the indispensable value for the method to work. Omitting it hides the intent. If I would encounter only `_io.bytes_strip_right(buf)`, I'd probably have to look in the method implementation _what on earth_ the method is stripping. Making it explicit like `_io.bytes_strip_right(buf, 0)` makes it clear.

You talk about _code readability_, but making parameters optional actually makes it worse in most cases.

kaitaistruct.py
```diff
@@ -125,6 +125,9 @@ def size(self):
     def read_s1(self):
         return KaitaiStream.packer_s1.unpack(self.read_bytes(1))[0]
 
+    def write_s1(self, data):
```
generalmimon, line 128: Please use the same order as in the existing [Java runtime library](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/serialization/src/main/java/io/kaitai/struct/KaitaiStream.java#L316-L387) (this means read methods and write methods coupled together, not this scattered mix). Again, maintainability and consistency purposes.

kaitaistruct.py
UserUnknownFactor: > Kaitai Struct runtime libraries are not primarily intended to be a _manual implementation helper_
This is not _manual implementation helper_ **per se** it is one **until** it's implemented on the compiler side. And it should be at some point because there is a serious need for native serialization support.

>There has to be some other reason why deviate from the established convention from Java.
There is no _convention_ from a single Java branch as it is in development and the only implementation. So anything should work.

Anyway it's a chicken and egg problem here. We need to write compiler support before runtime or runtime before compiler support. I decided to start from the runtime. 

Padding and termination can be done this way properly. That's because termination is a case of padding-right with a (single in case of ASCII C-strings) termination byte and it makes your whole first point invalid.

kaitaistruct.py
UserUnknownFactor: I see your point, thank you.

kaitaistruct.py
UserUnknownFactor: I don't get what you mean. Do you want me to move write functions away? So people rading the code couldn't compare them at a single glance and modify them without scrolling pages of other code just because some programmer in the single branch did so? Is there serious reason for this?

generalmimon: > I don't get what you mean.

It doesn't matter, I'll make it clear to you. The methods in the Python runtime library **MUST** appear in this order:

```java
// Reading
/// Integer numbers
//// Signed
abstract public byte readS1();
///// Big-endian
abstract public short readS2be();
abstract public int readS4be();
abstract public long readS8be();

///// Little-endian
abstract public short readS2le();
abstract public int readS4le();
abstract public long readS8le();

//// Unsigned
abstract public int readU1();

///// Big-endian
abstract public int readU2be();
abstract public long readU4be();
public long readU8be() {...}

///// Little-endian
abstract public int readU2le();
abstract public long readU4le();
public long readU8le() {...}

/// Floating point numbers
//// Big-endian
abstract public float readF4be();
abstract public double readF8be();

//// Little-endian
abstract public float readF4le();
abstract public double readF8le();


// Writing
/// Integer numbers
//// Signed
abstract public void writeS1(byte v);
///// Big-endian
abstract public void writeS2be(short v);
abstract public void writeS4be(int v);
abstract public void writeS8be(long v);

///// Little-endian
abstract public void writeS2le(short v);
abstract public void writeS4le(int v);
abstract public void writeS8le(long v);

//// Unsigned
abstract public void writeU1(int v);
///// Big-endian
abstract public void writeU2be(int v);
abstract public void writeU4be(long v);
abstract public void writeU8be(long v);

///// Little-endian
abstract public void writeU2le(int v);
abstract public void writeU4le(long v);
abstract public void writeU8le(long v);

/// Floating point numbers
//// Big-endian
abstract public void writeF4be(float v);
abstract public void writeF8be(double v);

//// Little-endian
abstract public void writeF4le(float v);
abstract public void writeF8le(double v);
```

> Do you want me to move write functions away?

Yes, I do. 

> So people reading the code couldn't compare them at a single glance and modify them without scrolling pages of other code

Let's make two comparisons which two functions are more similar to each other. 

1.
```py
def read_s1(self):
    return KaitaiStream.packer_s1.unpack(self.read_bytes(1))[0]
def write_s1(self, data):
    return self.write_bytes(KaitaiStream.packer_s1.pack(data))
```

2.
```py
def read_s1(self):
        return KaitaiStream.packer_s1.unpack(self.read_bytes(1))[0]
def read_s2be(self):
        return KaitaiStream.packer_s2be.unpack(self.read_bytes(2))[0]
```

Why would _people_ want to compare completely different functions `read_s1()` and `write_s1()` for which the greatest similarity is in the name?

UserUnknownFactor: Well if you say they must, they probably do. Because I feel like you are starting to harass me at this point. Why functions that operate on the same data type, use the same packer struct and do the opposite operations must be placed in a different parts of the code is beyond me.

UserUnknownFactor:
Okay, @generalmimon, if you implement the serialization support in .ksy compiler by yourself you can remove `write_bytes_term` and forget about the whole thing, how about it?

kaitaistruct.py
generalmimon: My point is actually quite simple: we can maintain the runtime libraries much easier when all of them use the same approach, logic and order of the functions, unless there's a very good reason not to do so (and different authors with different opinions and preferences is not among these reasons). One can simply split the screen with the Java runtime on the left and e.g. Python runtime on the right and see that there are exactly the same functions in both of runtimes.

If you believe that this established order doesn't make sense and you can come up with a much better one, feel free to open a [new issue](https://github.com/kaitai-io/kaitai_struct/issues/new) which will involve changing the order in all runtimes we have. There we can discuss it, but rather bring a bunch of solid arguments with you, because I'm still not convinced.

It's not possible to scatter chaos and confusion by using different approaches across runtimes for no reason. I can see that you're apparently not maintaining any such large project, so you just have to trust me, like it or not.

UserUnknownFactor:
This should be enough, right?

kaitaistruct.py
```diff
@@ -293,6 +385,21 @@ def read_bytes(self, n):
             )
         return r
 
+    def write_bytes(self, data, pad=0, padding=b'\0'):
```
generalmimon, line 388: You should follow the established approach from the Java runtime instead of devising a new one ([`ByteBufferKaitaiStream.java:532-555` on branch `serialization`](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/720ab6e6ac7b8875f4c56fe9e068f07f32b31d89/src/main/java/io/kaitai/struct/ByteBufferKaitaiStream.java#L532-L555)):

```java
    /**
     * Writes given byte array to the stream.
     * @param buf byte array to write
     */
    @Override
    public void writeBytes(byte[] buf) {
        bb.put(buf);
    }

    @Override
    public void writeBytesLimit(byte[] buf, long size, byte term, byte padByte) {
        int len = buf.length;
        bb.put(buf);
        if (len < size) {
            bb.put(term);
            long padLen = size - len - 1;
            for (long i = 0; i < padLen; i++)
                bb.put(padByte);
        } else if (len > size) {
            throw new IllegalArgumentException(
                    "Writing " + size + " bytes, but " + len + " bytes were given"
            );
        }
    }
```

The reason is that this method is already used by the compiler, and not only from the language specific part ([`JavaCompiler:834-835` on `serialization`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cfba0cccbdfa9b7ca77e498f9895ee6ecd2d7bfb/shared/src/main/scala/io/kaitai/struct/languages/JavaCompiler.scala#L834-L835)):

```scala
  override def attrBytesLimitWrite(io: String, expr: String, size: String, term: Int, padRight: Int): Unit =
    out.puts(s"$io.writeBytesLimit($expr, $size, (byte) $term, (byte) $padRight);")
```

but also from the common one ([`EveryWriteIsExpression:169` on `serialization`](https://github.com/kaitai-io/kaitai_struct_compiler/blob/cfba0cccbdfa9b7ca77e498f9895ee6ecd2d7bfb/shared/src/main/scala/io/kaitai/struct/languages/components/EveryWriteIsExpression.scala#L169)):

```scala
    attrBytesLimitWrite(io, expr, translator.translate(blt.size), term, padRight)
```

So when the Python serialization support in the compiler will be introduced, this `write_bytes` method that you wrote will have to be rewritten anyway.

Moreover, please follow the convention used in the rest of the Python library, that is: always accept a single byte as an integer number (`int`), instead of the `bytes` type in python. We've already discussed this at https://github.com/kaitai-io/kaitai_struct_python_runtime/pull/48#discussion_r465999296.

Ideally, just take the existing methods `writeBytes` and `writeBytesLimit` from Java runtime (shown above) and convert them to Python, with the exact same parameters and logic (if possible). You save yourself a lot of unnecessary work of inventing the new approach (and then eliminating bugs of this new logic), and we get a consistent system of methods across our runtime libraries.

kaitaistruct.py
```diff
@@ -293,6 +385,43 @@ def read_bytes(self, n):
             )
         return r
 
+    def write_bytes(self, data):
+        if data is None:
+            return
+        nb = len(data)
+        if nb > 0 and self._io.write(data) != nb:
+            raise Exception(
```
KOLANICH, line 393: Should be machine-readable.

kaitaistruct.py
```diff
@@ -293,6 +385,43 @@ def read_bytes(self, n):
...
+                "not all of %d bytes written" %
+                (nb,)
+            )
+
+    def write_terminator(self, term_byte):
+        self._io.write(KaitaiStream.packer_u1.pack(term_byte))
+
+    def write_padding(self, actual_size, size, pad_byte):
+        pad = size - actual_size
+        if pad > 0 and pad_byte is not None:
```
KOLANICH, line 403: >pad_byte is not None

what if `pad > 0` and `pad_byte is None`? Should it be an exception?

kaitaistruct.py
```diff
@@ -293,6 +385,43 @@ def read_bytes(self, n):
...
+            self._io.write(KaitaiStream.packer_u1.pack(pad_byte) * pad)
+
+    def write_bytes_term(self, data, size, term_byte, pad_byte):
+        if data is None or not size:
+            return
+        nb = len(data)
+        if nb > 0 and nb < size:
+            if self._io.write(data) != nb:
+                raise Exception(
```
KOLANICH, line 412: again, should be machine-readable.

kaitaistruct.py
```diff
@@ -320,9 +451,9 @@ def read_bytes_term(self, term, include_term, consume_term, eos_error):
     def ensure_fixed_contents(self, expected):
         actual = self._io.read(len(expected))
         if actual != expected:
-            raise Exception(
+            raise ValidationFailedError(
                 "unexpected fixed contents: got %r, was waiting for %r" %
-                (actual, expected)
+                (actual, expected), self._io, __file__
```
KOLANICH, line 456: They are still not machine readable. All the parameters should be put into exception object, not into the pregenerated str.

kaitaistruct.py
```diff
@@ -462,21 +590,21 @@ def __init__(self, expected, actual, io, src_path):
...
 class ValidationGreaterThanError(ValidationFailedError):
     """Signals validation failure: we required "actual" value to be
-    less than or equal to "max", but it turned out that it's not.
+    less than or equal to "maximal", but it turned out that it's not.
     """
-    def __init__(self, max, actual, io, src_path):
+    def __init__(self, maximal, actual, io, src_path):
         super(ValidationGreaterThanError, self).__init__("not in range, max %s, but got %s" % (repr(max), repr(actual)), io, src_path)
-        self.max = max
+        self.maximal = maximal
```
generalmimon, line 607: None of these changes are relevant to serialization. Please don't try to smuggle any unnecessary changes that only come from your personal opinions and preferences (and bring inconsistencies across runtime libraries, see https://github.com/kaitai-io/kaitai_struct_python_runtime/pull/48#discussion_r466304542) in a completely unrelated PR. For now, revert these changes and redo them in a separate PR if you want.

Oh, until now I haven't noticed that you opened this PR from the `master` branch of your fork repository, but it's still no excuse for doing such a big PR with many unrelated changes. Please remember this rule of thumb when making pull requests from your fork repository: **never commit directly on `master` of your fork** and **never open any pull requests from `master` of your fork**. Create a topic branch for every pull request you create. Read https://github.com/kaitai-io/kaitai_struct_doc/pull/38#issuecomment-653751666 for more info.

Using multiple topic branches allows you to create multiple pull requests at once, and thus you can keep the pull requests small and sticking to one particular topic. When you already opened this PR from `master`, keep it that way for now, but I still require that you stick to the topic of serialization which you have chosen.

Anyway, I see no reason here why rename `min` to `minimal` and `max` to `maximal`, all other runtime libraries use only `min` and `max`:

- [C++](https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/blob/d020781b96ea1e2fe7e0ecf47ff3ae3c829ccd31/kaitai/exceptions.h#L99-L143)
- [C#](https://github.com/kaitai-io/kaitai_struct_csharp_runtime/blob/1b66a9a85f39c52728893400a23504844cc78e34/KaitaiStruct.cs#L125-L161)
- [Java](https://github.com/kaitai-io/kaitai_struct_java_runtime/blob/c342c6e836ddba03c9ec33a59034e209bb04a976/src/main/java/io/kaitai/struct/KaitaiStream.java#L617-L641)
- [JavaScript](https://github.com/kaitai-io/kaitai_struct_javascript_runtime/blob/1fd3006373547792589310bc0746556bb17a697e/KaitaiStream.js#L760-L780)
- PHP - [`ValidationLessThanError`](https://github.com/kaitai-io/kaitai_struct_php_runtime/blob/f96d48c5df5cde7162dc3458bb059e6b02e7eab3/lib/Kaitai/Struct/Error/ValidationLessThanError.php#L6-L15), [`ValidationGreaterThanError`](https://github.com/kaitai-io/kaitai_struct_php_runtime/blob/f96d48c5df5cde7162dc3458bb059e6b02e7eab3/lib/Kaitai/Struct/Error/ValidationGreaterThanError.php#L6-L15)
- [Ruby](https://github.com/kaitai-io/kaitai_struct_ruby_runtime/blob/4cefe6f5700b138c15fc0d7633946591d16b7211/lib/kaitai/struct/struct.rb#L578-L598)

So please remember these sentences that I wrote you in https://github.com/kaitai-io/kaitai_struct_python_runtime/pull/48#discussion_r466304542:

> If you believe that this established order doesn't make sense and you can come up with a much better one, feel free to open a [new issue](https://github.com/kaitai-io/kaitai_struct/issues/new) which will involve changing the order in all runtimes we have.

> It's not possible to scatter chaos and confusion by using different approaches across runtimes for no reason.

This applies for these `min` and `max` parameter names as well.

It might look that these particular names don't matter, but it's rather about the principle. I'll keep stopping any [salami tactics](https://en.wikipedia.org/wiki/Salami_tactics) involving any bad practices already in its infacy.

kaitaistruct.py
```diff
@@ -293,6 +385,45 @@ def read_bytes(self, n):
...
 
+    def write_bytes(self, data):
+        if data is None:
+            return
+        nb = len(data)
+        if nb > 0 and self._io.write(data) != nb:
+            raise ValidationFailedError(
+                "not all of %d bytes written" %
+                (nb,), self._io, __file__
+            )
```
generalmimon, line 396: 1. The class `ValidationFailedError` is reserved for custom explicit format validations inside the `.ksy` spec (using the `valid` key, see https://github.com/kaitai-io/kaitai_struct/issues/435#issue-323332353), not for internal erroneous states of the runtime library. The point is that if the application code catches exception of type `ValidationFailedError`, the programmer can be sure that it's caused by a check performed by `contents` or `valid` key that he defined himself in the KSY format definition.

2. The `self._io` accessed inside the `KaitaiStream` class yields the original [`io.IOBase`](https://docs.python.org/3/library/io.html#io.IOBase)-derived stream resource, not the expected `KaitaiStream` object.

   Initializing will throw `AttributeError: '_io.BytesIO' object has no attribute 'pos'`, because the [`BytesIO` object](https://docs.python.org/3/library/io.html#io.BytesIO) has only `tell()` method, not `pos()`:

   https://github.com/kaitai-io/kaitai_struct_python_runtime/blob/4b974204a0d4d811ca5340d1b80acbfcface65b1/kaitaistruct.py#L444-L450

   You have to do just `self` instead of `self._io` to get the `KaitaiStream` object. Note that `self._io` is an instance of `KaitaiStream` only inside a `KaitaiStruct` class.

3. The `src_path` parameter is meant to be a YAML [XPath](https://cs.wikipedia.org/wiki/XPath) inside the source KSY spec pointing to a particular YAML key, not an absolute file path to the `.py` generated source code. The `.py` file path usually isn't much useful information.

   As you don't know the YAML key path here, you shouldn't use any class derived from [`KaitaiStructError`](https://github.com/kaitai-io/kaitai_struct_python_runtime/blob/4b974204a0d4d811ca5340d1b80acbfcface65b1/kaitaistruct.py#L425) at all.

If I were you, I would just raise the standard `Exception` for now in this case. If there's a desire to distinguish between Kaitai Struct errors and the other errors by type, we can think about introducing some exception hierarchy like in https://github.com/kaitai-io/kaitai_struct_php_runtime/pull/6, but please **not** in this PR. It's already quite congested, i.e. not really focusing on one topic.

kaitaistruct.py
```diff
@@ -320,10 +451,7 @@ def read_bytes_term(self, term, include_term, consume_term, eos_error):
     def ensure_fixed_contents(self, expected):
         actual = self._io.read(len(expected))
         if actual != expected:
-            raise Exception(
-                "unexpected fixed contents: got %r, was waiting for %r" %
-                (actual, expected)
-            )
+            raise ValidationNotEqualError(expected, actual, self._io, __file__)
```
generalmimon, line 454: Please revert this. The `ensure_fixed_contents` method is left here only for backward compatibility with generated code by 0.8 compiler and lower (the 0.9 compiler-generated code no longer uses this method), and thus it makes a little sense why raise a more specific type here. It can actually break the compatibility, because you completely changed the error message by doing this (if some KS user does string matching on the error message to distinguish this exception among others, it probably won't work anymore).

And unsurprisingly, it doesn't have to do anything with serialization again.

PR #49, opened 2020-08-18
Add API_VERSION constant storing the runtime version as a tuple of ints

dgelessus:
Part of one of the potential fixes for kaitai-io/kaitai_struct#804.


dgelessus:
Since this topic came up again in #62 - any objections to merging this (followed by kaitai-io/kaitai_struct_compiler#208)? I don't remember all of the details of the long discussion in kaitai-io/kaitai_struct#804, but IIRC there were no strong objections to this solution.

dgelessus:
Actually, I need to update this first to bump all of the version numbers. I created this PR before the release of KS 0.9, so all of the comments still say 0.9 and not 0.10.

KOLANICH:
IMHO the symbol should either start with underscore or have `KAITAI` in the beginning of the name. And `__version__` string should be generated automatically.

dgelessus:
> IMHO the symbol should either start with underscore or have `KAITAI` in the beginning of the name.

Why do you think the name should be changed this way?

An underscore would be misleading. `API_VERSION` is not an internal variable - quite the opposite, it will be used by ksc-generated code, so it has to be kept stable in future versions of the runtime. An extra `KAITAI` prefix would be redundant IMO, because the attribute is already namespaced under the `kaitaistruct` module, so there should be no confusion about which API version it's referring to.

> IMHO the symbol should either start with underscore or have `KAITAI` in the beginning of the name. And `__version__` string should be generated automatically.

I don't think we can do that. The `__version__` value needs to be a string literal in the source code, so that setuptools can statically extract the version number from the code (for the `attr: kaitaistruct.__version__` directive in setup.cfg).

dgelessus:
Rebased on top of #64 to update the Kaitai Struct version number.

KOLANICH:
>so that setuptools can statically extract the version number from the code (for the attr: kaitaistruct.__version__ directive in setup.cfg).

I am not sure it does it statically, for static extraction there is `read_version` package.

dgelessus:
> I am not sure it does it statically, for static extraction there is `read_version` package.

[setuptools 46.4.0](https://setuptools.pypa.io/en/latest/history.html#v46-4-0) and later will parse attribute values statically if possible. If that fails, it still falls back to importing the module, but I would like to avoid that if possible.

generalmimon (approved):
@dgelessus LGTM, thanks!

PR #54, opened 2020-11-18
Add more PyPI badges

generalmimon:
I've already added a badge showing the `kaitaistruct` package version at PyPI, check out the [readme](https://github.com/kaitai-io/kaitai_struct_python_runtime#readme):

PyPI

But there are other badges at https://shields.io/ that show additional info that can be useful for users. I added all of them that I consider potentially reasonable or relevant for this project, and I'd like to hear from Python experts what information is not relevant.

Another thing I'd like to know is whether there are some pages at PyPI or other websites that the badges could link to when clicked. The point is to provide some proof or basis for the information that the badge claims - for example, when the badge claims

PyPI - Downloads

..., it would be nice to link to a PyPI page where one can look at a more detailed graph with the number of downloads over time. However, it seems that PyPI doesn't provide it - then isn't there some third-party service that does?

For example, in the JavaScript ecosystem, there is [npmtrends.com](https://www.npmtrends.com/kaitai-struct-compiler), which shows exactly this for `npm` packages (but it's not affiliated to [npmjs.com](https://www.npmjs.com/) authors or GitHub in any way - it's created [by a guy John Potter](https://github.com/johnmpotter/npm-trends)). You can even put more packages in the graph and compare which one is/was more popular over time - e.g. [kaitai-struct-compiler vs. jbinary](https://www.npmtrends.com/kaitai-struct-compiler-vs-jbinary).

If there's no specific for that measure, that's fine: it just won't be clickable. I'd like to avoid linking all badges to https://pypi.org/project/kaitaistruct/ when the information is not there - that gives the reader false hope that he can learn something more about the particular statement of the badge, but he'd be actually sent to a general `kaitaistruct` PyPI package homepage where he didn't want to go (if he did, he'd click the [PyPI](https://pypi.org/project/kaitaistruct/)).

Cc @GreyCat, @dgelessus, @arekbulski, @cugu

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,8 @@
 # Kaitai Struct: runtime library for Python
 
 [![PyPI](https://img.shields.io/pypi/v/kaitaistruct)](https://pypi.org/project/kaitaistruct/)
-![PyPI - Format](https://img.shields.io/pypi/format/kaitaistruct)
-![PyPI - Wheel](https://img.shields.io/pypi/wheel/kaitaistruct)
 ![PyPI - Python Version](https://img.shields.io/pypi/pyversions/kaitaistruct)
-![PyPI - Implementation](https://img.shields.io/pypi/implementation/kaitaistruct)
-![PyPI - Downloads](https://img.shields.io/pypi/dd/kaitaistruct)
+[![PyPI - Downloads](https://img.shields.io/pypi/dd/kaitaistruct)](https://pypistats.org/packages/kaitaistruct)
 
 This library implements Kaitai Struct API for Python.
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,11 @@
 # Kaitai Struct: runtime library for Python
 
 [![PyPI](https://img.shields.io/pypi/v/kaitaistruct)](https://pypi.org/project/kaitaistruct/)
+![PyPI - Format](https://img.shields.io/pypi/format/kaitaistruct)
+![PyPI - Wheel](https://img.shields.io/pypi/wheel/kaitaistruct)
+![PyPI - Python Version](https://img.shields.io/pypi/pyversions/kaitaistruct)
+![PyPI - Implementation](https://img.shields.io/pypi/implementation/kaitaistruct)
+![PyPI - Downloads](https://img.shields.io/pypi/dd/kaitaistruct)
 
 This library implements Kaitai Struct API for Python.
 


dgelessus:
For download statistics there's pypistats: https://pypistats.org/packages/kaitaistruct

These stats aren't always very accurate or useful though - see [this explanation](https://packaging.python.org/guides/analyzing-pypi-package-downloads/#background) in the Python packaging user guide.

About the badges: I think the one with the supported Python versions is useful and the download count is nice to have. But I don't think the "format: source" and "wheel: no" badges are very useful or interesting, because that's not something you usually care about as a user, and providing wheels is so common that it's not really worth pointing out IMO. I don't have an opinion on the "implementation" badge - I don't really use PyPy, so I don't know how useful that information is.

I don't think there's anything useful to link to for the other badges, except for the download stats. The PyPI page contains the information shown by all the other badges - the Python version and implementation info is listed under "Classifiers" in the left sidebar, and the available distribution formats (source, wheel) can be seen under "Download files".

generalmimon:
> For download statistics there's pypistats: https://pypistats.org/packages/kaitaistruct

Thanks for sharing this, I'll link to it.

> These stats aren't always very accurate or useful though - see [this explanation](https://packaging.python.org/guides/analyzing-pypi-package-downloads/#background) in the Python packaging user guide.

Hmm, I've read their reasons, but they didn't actually convince me, and I'd really doubt the label _**highly** inaccurate_, if I look at the numbers the `kaitaistruct` package has. This undermining of value of the download stats seems to be more of an [epistemological skepticism](https://en.wikipedia.org/wiki/Philosophical_skepticism#Epistemological_skepticism), i.e. _certainty in knowledge is impossible_. And this idea can be applied to absolutely everything.

And yeah, of course that the number of users of any project doesn't correlate to its quality, but that doesn't imply it's not useful to know.

> But I don't think the "format: source" and "wheel: no" badges are very useful or interesting, because that's not something you usually care about as a user, and providing wheels is so common that it's not really worth pointing out IMO. I don't have an opinion on the "implementation" badge - I don't really use PyPy, so I don't know how useful that information is.

Thanks, I will remove them. I also thought that they're not so interesting, but I don't know much about the Python ecosystem, so I wasn't sure.

And yeah, implementation is also a weird thing. I assume most Python code that works in cpython also works in PyPy anyway, so the value of the information is disputable. Moreover, we don't actually run our [tests](https://github.com/kaitai-io/kaitai_struct_tests) on PyPy, so we don't even know if it works there (in the spirit of the _epistemological skepticism_).

generalmimon:
Oh my ..., why doesn't GitHub allow embedding images without linking to somewhere? No matter how hard I try, clicking on the badge "Python versions" always opens the badge in a new tab.

I guess I'll have to explicitly link to somewhere then... @dgelessus Can you think of a good destination?

dgelessus:
Yeah, I think that's simply GitHub's default behavior for images, which is useful in most cases, but not here :) I would simply link the versions badge to the PyPI page as well, because the version information can be found there.

PR #56, opened 2020-12-17
Fixed some pylint warnings

KOLANICH:


kaitaistruct.py
```diff
@@ -451,19 +453,19 @@ class ValidationLessThanError(ValidationFailedError):
     """Signals validation failure: we required "actual" value to be
     greater than or equal to "min", but it turned out that it's not.
     """
-    def __init__(self, min, actual, io, src_path):
-        super(ValidationLessThanError, self).__init__("not in range, min %s, but got %s" % (repr(min), repr(actual)), io, src_path)
-        self.min = min
+    def __init__(self, minBound, actual, io, src_path):
```
generalmimon, line 456: I don't like the new name. The old `min` is clear enough.

And BTW, you have to keep the naming convention consistent all the time. You just can't have two semantically equivalent identifiers `minBound` and `src_path` on the same line. Use the `snake_case` for parameter names as the rest of the code does.

generalmimon (changes requested):

dgelessus (commented):
Honestly I'm not a fan of most of these changes...

Renaming the `min` and `max` parameters makes some sense, to avoid conflicting with the built-in `min` and `max` functions.

I don't see the benefit of unwrapping *every* `else` branch where the corresponding `if` ends in a `return`. There are some cases where this change makes sense and makes the code more readable, for example when the `if` branch is an obvious "error" case where the code is exited early. But in other cases this change makes the code *less* readable IMO, for example in the case of `if PY2: ... else: ...`, where both branches of the code are equally long and there's no obvious "success" and "error" branch.

The remaining changes are just disabling a bunch of pylint checks. I would argue that it's not really worth using a linter if you have to disable most of its checks and it shows almost no warnings that are actually useful...

To be clear, I have nothing against linters or code style checking in general. I just don't like tools where most of the warnings are incorrect or unhelpful, and where you need to disable/suppress many checks before you get actually useful output.

KOLANICH:
>I don't see the benefit of unwrapping every else branch where the corresponding if ends in a return

It maps to python bytecode.

```python
In [1]: def a(): 
   ...:     if c(): 
   ...:         return 1 
   ...:     else: 
   ...:         return 2 
   ...:                                                                                                            

In [2]: import dis                                                                                                 
In [3]: dis.dis(a)                                                                                                 
  2           0 LOAD_GLOBAL              0 (c)
              2 CALL_FUNCTION            0
              4 POP_JUMP_IF_FALSE       10

  3           6 LOAD_CONST               1 (1)
              8 RETURN_VALUE

  5     >>   10 LOAD_CONST               2 (2)
             12 RETURN_VALUE
             14 LOAD_CONST               0 (None)
             16 RETURN_VALUE

In [4]: def b(): 
   ...:     if c(): 
   ...:         return 1 
   ...:     return 2 
   ...:                                                                                                            

In [5]: dis.dis(b)                                                                                                 
  2           0 LOAD_GLOBAL              0 (c)
              2 CALL_FUNCTION            0
              4 POP_JUMP_IF_FALSE       10

  3           6 LOAD_CONST               1 (1)
              8 RETURN_VALUE

  4     >>   10 LOAD_CONST               2 (2)
             12 RETURN_VALUE
```

kaitaistruct.py
```diff
@@ -367,7 +369,7 @@ def process_rotate_left(data, amount, group_size):
         anti_amount = -amount & mask
 
         r = bytearray(data)
-        for i in range(len(r)):
+        for i in range(len(r)):  # pylint:disable=consider-using-enumerate
```
generalmimon, line 372: Hmm, using `enumerate()` as pylint suggests actually looks like a good idea. It results in a more Pythonic code and it is apparently faster as well - see my [simple benchmark](https://gist.github.com/generalmimon/31612fa008753ef0943558e1d8104496):

```console
$ python --version
Python 3.9.6

$ python bench-process-rotate-w-enumerate.py
enumerate loop: [7.436, 7.548, 7.598, 7.622, 7.685]
original "for i in range(len(r))" loop: [8.412, 8.423, 8.448, 8.451, 8.465]

$ python bench-process-rotate-w-enumerate.py
original "for i in range(len(r))" loop: [8.179, 8.190, 8.221, 8.330, 8.397]
enumerate loop: [7.460, 7.500, 7.532, 7.539, 7.555]

$ python2 --version
Python 2.7.18

$ python2 bench-process-rotate-w-enumerate.py
('enumerate loop:', '[24.086, 24.571, 24.621]')
('original "for i in range(len(r))" loop:', '[31.043, 31.066, 31.204]')

$ python2 bench-process-rotate-w-enumerate.py
('original "for i in range(len(r))" loop:', '[30.531, 30.922, 31.063]')
('enumerate loop:', '[24.746, 24.912, 25.190]')
```

(I tried both orders of the functions, because sometimes I couldn't shake a feeling from benchmarks in other languages that the execution time of a case that runs first/last is skewed. But specifically [`timeit`](https://docs.python.org/3/library/timeit.html) seems to avoid this issue, because I don't see any bias here and the results are printed to the console at the same time, which makes an impression to me that [`timeit`](https://docs.python.org/3/library/timeit.html) somehow runs both tests at the same time, which sounds like it would really fix the problem.)

generalmimon (approved):
@KOLANICH Looks good, thanks!

After merging, I'll make a few more related edits.

generalmimon:
Maybe we can consider adding `pylint` to the CI check. Currently we only use `pycodelint`, but `pylint` reports some issues that are arguably more important than where should be 2 newlines instead of 1. That is good to know as well, but having only that _instead of_ a linter which suggests improvements with an impact on functionality is a bit... ironic.

KOLANICH:
>Maybe we can consider adding pylint to the CI check.

Maybe. But IMHO it may not worth to delay the CI by pylint checks. GH Actions is done not very rationally. They run own pretty bloated Docker image, and if a user has chosen to base his pipeline on Docker, his Docker image is run within GH Docker image instead of providing a way for a user to just add a few layers above GH Docker image and use it directly over their infrastructure. So GH has decreased its value of using Docker images as a cache for dependencies. So the most of actions install dependencies from scratch, it takes some time, usually greater than pull a small layer and apply it over a image already used by multiple users in multiple pipelines. Every delay in CI makes usage of the CI less worthy.

PR #57, opened 2020-12-17
Just some reformatting using autopep8 + black (some fixes were reverted)

KOLANICH:


generalmimon:
> _No description provided._

I don't see any purpose of these changes and I don't think they are reasonable.

Issue #69, opened 2022-07-09
PyPI project page does not link to GitHub repo


generalmimon:
I've just noticed that https://pypi.org/project/kaitaistruct/ doesn't contain any backlink to the source repository, i.e. https://github.com/kaitai-io/kaitai_struct_python_runtime. The "Homepage" in Project links leads to https://kaitai.io:

![image](https://user-images.githubusercontent.com/47499687/178085954-11e0c021-795b-437e-a1f3-4eb7cde93dd6.png)

..., which I guess is fine because it's indeed the homepage of the _project_, but the repo with source code of the package at PyPI is not directly reachable.

I wonder how it can be expressed in `setup.cfg` - it looks like [`project_urls`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) is intended for that purpose?

> List additional relevant URLs about your project. This is the place to link to bug trackers, source repositories, or where to support package development. The string of the key is the exact text that will be displayed on PyPI.

`project_urls` needs [`setuptools >= 38.3.0`](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html#metadata), which is fine, because we already require an even later version (because of the `long_description_content_type` we're using):

https://github.com/kaitai-io/kaitai_struct_python_runtime/blob/c29609940a4cee756c35b5d6d429ba00aa4ec99f/setup.cfg#L37-L38

dgelessus:
Yes, `project_urls` is the correct thing to use here. See for example:

https://github.com/beeware/rubicon-objc/blob/d4533001eee0c7349ebf0007993e9489ba948fc6/setup.cfg#L5-L9

PR #71, opened 2022-09-16
setup.cfg: add `metadata.project_urls`

generalmimon:
Fix https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/69

According to a project example https://pypi.org/project/PyScaffold/4.3/ + https://github.com/pyscaffold/pyscaffold/blob/v4.3/setup.cfg#L7-L15, the "Homepage" link pulled from `metadata.url` apparently stays even if I add `project_urls`, so all links at PyPI resulting from this change should be (I'm not sure about their order, but again links on https://pypi.org/project/PyScaffold/4.3/ seem to be sorted alphabetically except the implicit "Homepage" link at the beginning):

- [Homepage](https://kaitai.io)
- [Documentation](https://doc.kaitai.io/)
- [Gitter](https://gitter.im/kaitai_struct/Lobby)
- [Source](https://github.com/kaitai-io/kaitai_struct_python_runtime)
- [Tracker](https://github.com/kaitai-io/kaitai_struct_python_runtime/issues)

Currently (see https://pypi.org/project/kaitaistruct/0.10/) the project links are just:

- [Homepage](https://kaitai.io)


setup.cfg
```diff
@@ -4,6 +4,11 @@ version = attr: kaitaistruct.__version__
 author = Kaitai Project
 author_email = greycat@kaitai.io
 url = https://kaitai.io
+project_urls =
+    Source = https://github.com/kaitai-io/kaitai_struct_python_runtime
+    Tracker = https://github.com/kaitai-io/kaitai_struct_python_runtime/issues
```
generalmimon, line 9: We can also add some "Documentation" link, but I'm not sure if it should link to https://doc.kaitai.io/, https://doc.kaitai.io/lang_python.html or somewhere else.

KOLANICH:
IMHO it'd be a bit redundant to add the links `source code` and `issue tracker` since everyone knows that source code hostings (GitHub currently) are used for that. So IMHO 1 homepage link to this repo is enough. The links to the docs ... you don't have to limit oneself. Proposal: 1 link to the docs for python runtime, and another one to the guide to KS language.

KOLANICH:
Also: `setup.cfg` is deprecated, convert it to `pyproject.toml` using `ini2toml` tool.

dgelessus (approved):
The list of links looks good to me. I personally wouldn't have added a Twitter link here (I think it's not all that relevant when coming from PyPI), but I have no problem with keeping it if the list doesn't become too long.

A link to the docs sounds good, but I have no strong opinion about linking to the main docs or the Python notes (the reader can always click through to the other docs if needed).

dgelessus:
> Also: `setup.cfg` is deprecated,

From where do you have that information? The [setuptools docs for `setup.cfg`](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) don't mention anything about it being deprecated. All I can find is [pypa/setuptools/milestone/7](https://github.com/pypa/setuptools/milestone/7) about *eventually* deprecating `setup.cfg`, but that hasn't happened yet. The setuptools maintainers also mentioned in [multiple](https://github.com/pypa/setuptools/issues/1688#issuecomment-471271447) [places](https://github.com/pypa/setuptools/issues/1160#issuecomment-549054957) that the deprecation period would probably be "forever", so there's no rush.

> convert it to `pyproject.toml` using `ini2toml` tool.

That's still not an option. [As before](https://github.com/kaitai-io/kaitai_struct_python_runtime/pull/39#pullrequestreview-937134754), `pyproject.toml`-based configuration requires setuptools 61.0.0, which requires Python 3.7, and we still supports version before that.

generalmimon:
@dgelessus:

> The list of links looks good to me. I personally wouldn't have added a Twitter link here (I think it's not all that relevant when coming from PyPI)

I agree that it's not that relevant for the Python runtime, and it's not clear what a Python user would find there.

generalmimon:
@dgelessus:

> A link to the docs sounds good, but I have no strong opinion about linking to the main docs or the Python notes (the reader can always click through to the other docs if needed).

That's true, so I think the main docs are fine.

PR #73, opened 2022-12-09
add support local implement for micro-python

alingse:
I was using kaitaistruct to process some binary data with a micro-python low level device.
it does not have the module `itertools`, since this package only use `cycle` for make a loop iter on the bytearray

so i add an `local_cycle` implement for the env without itertools 

and i add a test for it, it was run with a pytest

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,3 +9,8 @@ jobs:
       - uses: actions/checkout@v1
       - name: Python Style Checker
         uses: andymckay/pycodestyle-action@0.1.3
+
+      - name: Run Pytest
+        uses: fylein/python-pytest-github-action@v1
+        with:
+          args: pytest
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -9,13 +9,19 @@ except ImportError:
     SEEK_END = 2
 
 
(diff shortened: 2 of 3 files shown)


alingse:
i found the new version have the implement  https://github.com/micropython/micropython-lib/blob/master/python-stdlib/itertools/itertools.py#L7-L19

but i still want it.

KOLANICH:
1. 
```python
    try:
        len(p)
    except TypeError:
        # len() is not defined for this type. Assume it is
        # a finite iterable so we must cache the elements.
        cache = []
        for i in p:
            yield i
            cache.append(i)
        p = cache
```

is not needed, in this lib `cycle` is always used on types having `__len__`.

2. You should define the function if the import fails.
3. Well, the copied impl specific to the case that the key is a collection and non-empty (so `while True` should fit) should have lower overhead than the generic one, so it may make sense to use it instead of the imported one if it is not a builtin
4. it may make more sense to make `cycle` a builtin of `micropython` than bring it here in the form of python code.
5. probably you should collapse all the commits into one.

KOLANICH:
6. This PR brings a third-party GitHub Docker action. It is inacceptable.

PR #80, opened 2024-09-15
Adopt a system of exceptions derived from KaitaiStructError

### LINKED ISSUE #40: Define a custom exception class instead of raising generic exceptions
Currently, when a Kaitai Struct parser detects that the input data is invalid, it can raise a variety of exceptions: `ValueError`, `EOFError`, generic `Exception`, and possibly some others. This makes it very difficult for callers to specifically catch parse errors.

It would be helpful to have a custom exception class (e. g. `kaitaistruct.ParseError`) that is consistently raised for all invalid input data. This would allow the calling code to catch and handle that exception specifically, without catching other unrelated errors by accident.

### LINKED ISSUE #41: read_bytes_term() should raise EOFError instead of Exception
`read_bytes_term()` should raise EOFError instead of Exception to be consistent with `read_bytes()`.

Btw. it's funny that found this issue 1h later than the person that has reported #40 

Even if #40 is implemented using a custom class, I'd suggest having both `read_bytes_term()` and `read_bytes()` raise the same type of error.

My use case is that I'm trying to detect when the packet that I'm parsing has been truncated so it would be helpful to be able to just do `except EOFError` (or `except kaitaistruct.EOFError`) instead of having to parse the exception message.

generalmimon:
* Resolves https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/40
* Resolves https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/41

As explained in https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/40, this makes it easy to handle to all errors caused by invalid input data by using `kaitaistruct.KaitaiStructError` in a `try..except` statement. Three new exception types were added: `InvalidArgumentError`, `EndOfStreamError` and `NoTerminatorFoundError`. All changes to raised exceptions in this commit should be backward compatible, as we are only moving to subclasses of previously raised exceptions. `NoTerminatorFoundError` is a subclass of `EndOfStreamError` to address the suggestion in https://github.com/kaitai-io/kaitai_struct_python_runtime/issues/41.

Note that the `process_rotate_left` method could only raise `NotImplementedError` if someone called it manually (because KSC-generated parsers hardcode `group_size` to `1`, see https://github.com/kaitai-io/kaitai_struct_compiler/blob/c23ec2ca88d84042edba76f70c1f003d062b7585/shared/src/main/scala/io/kaitai/struct/languages/PythonCompiler.scala#L211), so it makes no sense to raise an exception derived `KaitaiStructError` (it's a programmer error, not a user input error). Most of our runtime libraries in other languages don't even have this `group_size` parameter, and if they do (C#, Java, Ruby), they also throw the equivalent of `NotImplementedError` (except the JavaScript runtime, which throws a plain string, which is _possible_ in JS but considered bad practice, so we should fix this).

### DIFF
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -739,14 +739,25 @@ class KaitaiStream(object):
 
     def write_bytes_limit(self, buf, size, term, pad_byte):
         n = len(buf)
+        # Strictly speaking, this assertion is redundant because it is already
+        # done in the corresponding _check() method in the generated code, but
+        # it seems to make sense to include it here anyway so that this method
+        # itself does something reasonable for every set of arguments.
+        #
+        # However, it should never be `false` when operated correctly (and in
+        # this case, assigning inconsistent values to fields of a KS-generated
+        # object is considered correct operation if the user application calls
+        # the corresponding _check(), which we know would raise an error and
+        # thus the code should not reach _write() and this method at all). So
+        # it's by design that this throws AssertionError, not any specific
+        # error, because it's not intended to be caught in user applications,
+        # but avoided by calling all _check() methods correctly.
+        assert n <= size, "writing %d bytes, but %d bytes were given" % (size, n)
+
         self.write_bytes(buf)
         if n < size:
             self.write_u1(term)
-            pad_len = size - n - 1
-            for _ in range(pad_len):
-                self.write_u1(pad_byte)
-        elif n > size:
-            raise ValueError("writing %d bytes, but %d bytes were given" % (size, n))
+            self.write_bytes(KaitaiStream.byte_from_int(pad_byte) * (size - n - 1))
 
     # endregion
 
--- a/kaitaistruct.py
+++ b/kaitaistruct.py
@@ -128,6 +128,9 @@ class KaitaiStream(object):
         return self._io.tell() >= self.size()
 
     def seek(self, n):
+        if n < 0:
+            raise InvalidArgumentError("cannot seek to invalid position %d" % (n,))
+
         if self.bits_write_mode:
             self.write_align_to_byte()
         else:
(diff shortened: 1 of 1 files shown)


kaitaistruct.py
```diff
@@ -128,6 +128,9 @@ def is_eof(self):
         return self._io.tell() >= self.size()
 
     def seek(self, n):
+        if n < 0:
+            raise InvalidArgumentError("cannot seek to invalid position %d" % (n,))
```
generalmimon, line 132: For the record, this exception was already thrown by the [`io.IOBase.seek()`](https://docs.python.org/3/library/io.html#io.IOBase.seek) method in the Python standard library. The exception type and message text and depended on which underlying stream was used. When using [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO), a `ValueError("negative seek value %zd")` was thrown (see [`Modules/_io/bytesio.c:659-660`](https://github.com/python/cpython/blob/401fff7423ca3c8bf1d02e594edfd1412616a559/Modules/_io/bytesio.c#L659-L660)). But when using a regular file, I'm getting `OSError: [Errno 22] Invalid argument` on both Windows and Linux.

So it seems better to do this check ourselves. Also a negative seek position can easily be a result of invalid input data (and thus the exception should be a subclass of `KaitaiStructError`) - imagine a situation like this:

```ksy
seq:
  - id: ofs_foo
    type: s4
instances:
  foo:
    pos: ofs_foo
    type: u1
```

If `ofs_foo` is negative and we attempt to read `foo`, it's a clear parse error.

generalmimon:
I'm going to merge this now, but feel free to comment on this PR even after the merge.

